PR #16, opened 2018-06-08
Decimal weight

lobneroO:
Changed the weight to support decimal numbers.

First commit as a try out


davidhealey:
Weight already supports floating point numbers, are you using the latest version?

PR #19, opened 2018-06-15
Theme selector as well as AMOLED and night theme added

maste9:
Changing of theme on the fly works well. Unfortunately the app seems not to remember on app's startup and I don't get why. Maybe you find what I did wrong?
The new stylesheets should later be optimized but the concept should proof for the moment.


davidhealey:
Thanks, I'll try it out and see what needs doing

PR #24, opened 2018-07-23
Create locale-de.json

achimwagenknecht:
My son translated this into german.


davidhealey:
Excellent, thank you!

achimwagenknecht:
Thanks for accepting and merging the file so quickly. I would love to see the translation on Google Play and F-Droid. I'm not familiar with the procedure, will it be included in the next release automatically?

davidhealey:
I have to do the Play Store release manually (which is why it's a few versions behind). The F-Droid one will update automatically once I issue a release on github. I'm currently working on a massive UI rewrite which will form Waistline v2.0. So I will probably do the next release once that version is finished, hopefully won't be too long, and it will include your translation as well.

achimwagenknecht:
Thank you very much for your great work. I have been looking for an open source calories counter for years!

maste9:
Hmm ... still waiting for 1.0.3 and 1.0.4 on F-droid actually. Is there anything wrong with the process maybe? Or does it always take weeks to be published there because of reviewing?

davidhealey:
Hi, I have good news and slightly less good news. The good news is that version 2 of Waistline with many added features will be released tomorrow or the next day. The other news it that version 2 was a massive rewrite and there is not only a load of new strings in the language file but I also changed the formatting to make it more manageable. Unfortunately this means the language files for version 1 are not compatible.

achimwagenknecht:
OK, where are the new language files?

Issue #69, opened 2018-09-26
F-droid updates


grrrrr:
I am not sure if this is something you have control or influence over David but 2.3.4 is not available on F-Doid still.

https://f-droid.org/wiki/page/com.waist.line#2.3.4

https://f-droid.org/wiki/page/com.waist.line/lastbuild_20304

davidhealey:
I have no control over it. F-Droid builds the releases from source automatically, I'm not sure what the interval is between builds but it seems to be a bit random.

relan:
2.3.4 is not on F-Droid because of #66.

davidhealey:
Oh, any idea how to resolve it?

relan:
Nope. I never had any experience with Cordova.

davidhealey:
hmmm I'll see what I can do

relan:
F-Droid's build recipe was relying on Gradle. Now that build.gradle is gone, it doesn't know how to find and build updates.

davidhealey:
I see, so I need to push the platforms directory, no problemo. Thank you

relan:
It was a very bad idea to push all *.bin, *.lock, *.class and other binary files.

davidhealey:
Tell me more

relan:
Could you amend the last commit and push it with `-f`?

relan:
Whatever is needed to make `gradle assemblyRelease` succeed. I guess it will be OK to remove all `.gradle` and `build` directories.

davidhealey:
Ok thanks. I don't know anything about Android development, I let Cordova handle it, so I appreciate your input very much.

davidhealey:
Ok I think I've done that successfully, seems like those files are no longer in the git repo.

relan:
Thanks! There are still some leftovers (as of 142a974):

```
$ find . -type d -name build
./platforms/android/CordovaLib/build
```

This is _not_ a showstopper for an F-Droid release though. Added 2.4.2 in https://gitlab.com/fdroid/fdroiddata/commit/717d40457fc3f4e062142fe1a66f623c62c13082.

davidhealey:
Thank you, I'll remove those files before the next update

PR #91, opened 2018-11-12
Cleanup

corngood:
- remove all cordova generated files: node_modules, platforms/, plugins/, and update .gitignore
- remove executable flag from source files
- update android platform and browsersync plugin (to fix a build error)
- add travis CI script
- add emacs .dir-locals.el to set 2-space indentation for javascript

@davidhealey 

Could you try this out when you get a chance and make sure everything still works the way you expect?  You'll need to run `cordova prepare` to repopulate the generated files.

Also, if you enable Travis CI on this repo, it should work.  If you have another preferred CI system, let me know.


davidhealey:
Well something's not right

```
cordova prepare browser
Unable to load PlatformApi from platform. Error: Cannot find module '/media/dave/Work 2/Android Studio Projects/waistline/platforms/browser/cordova/Api.js'
Unhandled "error" event. (The platform "browser" does not appear to be a valid cordova platform. It is missing API.js. browser not supported.)
```

I don't have time to work this out at the moment though.

corngood:
Did you try `cordova prepare`?  It should do both android and browser because they are both in the config.xml.

I have seen this type of error before, but cleaning the workspace and preparing always fixed it.

corngood:
@davidhealey I'm just trying the browser build for the first time.  I did `cordova build browser`, which completed with no output, and then opened `platforms/browser/www/index.html`.  Is that the right thing to do?

The app partially works in firefox.  I could search for food items on OFF, but the add food page is pretty broken and I couldn't find a way to actually add the item (there's no checkmark button like on android).

I'm just wondering if this is how it normally works, if I built it wrong, or if I broke it with my changes.

davidhealey:
I think something is broken but I'll need to check further. I do all my development in the browser (followed by final debugging on my Android devices).

You'll want to have the browser live refresh plugin installed (that might be something that was in the node or platforms folder I'm not quite sure), then you cd to the project folder and use this command **cordova run browser -- --live-reload**. This will start a virtual server which you can connect to in your browser (I use firefox) by going to http://localhost:8000/. Then it will run just like the app (I hit F12 and switch to responsive design mode to simulate a mobile device).

corngood:
Ok, thanks.  I tried that method of running and it seems to work and reload files properly, at least on my system.  I'm using cordova 8.1.2 btw.

davidhealey:
Ok, I'll give it another run later and see if it works for me. Did you do any other steps?

corngood:
No, just `cordova prepare`, `cordova run browser -- --live-reload`.  There are still some problems with the browser version though (on mine anyway):

- missing icons for the menu button, back, accept, etc.
- import/export doesn't seem to work

corngood:
@davidhealey do you mind enabling Travis CI for the repo?  Then we can start getting build results on PRs.

davidhealey:
I don't mind, but I have no idea what it is or how I do it, please educate me :)

corngood:
@davidhealey It's a free continuous integration system which integrates with github.  I think it's the most popular one by a long way.  

You just need to go to https://travis-ci.org/ and authenticate with your GH account, and then enable testing of this repo.

Then you'll get the equivalent of this, which I enabled for my fork: https://travis-ci.org/corngood/waistline.  On GH you get notifications of test status on PRs, and badges for commits like on: https://github.com/corngood/waistline/commits/timezones.

davidhealey:
Thanks. I've enabled it now on the Repo, let me know if I need to do anything else to get it running.

corngood:
That should be fine.  Next time you push something, or someone creates a PR, it should do builds, notify on failure, etc.

So far it's only testing that an apk can build, but it gives us somewhere to add tests e.g. js linting, automated testing.

davidhealey:
I'm going to re-add the platforms >> android folder to the git repo because it turns out F-Droid is relying on it to create updates.

corngood:
It doesn't sound right that it would need build artifacts in the repo, but I don't know anything about fdroid packaging.  Is it documented somewhere?  How do you configure it?

corngood:
Actually I think I found the relevant portion of the docs:

> Special note on Cordova/Phonegap/HTML-Apps:
>
> We cannot build cordova apps directly, but recent version allows you to export the platform-specific code that can be build standalone using “gradle”. So for now we require this code to be present and up-to-date in the source repository.

That's unfortunate.

davidhealey:
More detail here - https://github.com/davidhealey/waistline/issues/69#issuecomment-455290095 

Seems the build folders aren't needed

PR #94, opened 2018-11-12
Switch diary dates to UTC

### LINKED ISSUE #52: Dates keep reseting and changing
I'm having an issue on my  device running android  8.1 where the default day in the diary is never the day it currently is. It will often reset to the day before or the day before that.  

I can manually select today's date,  but after I add the food or go back to the diary later in the day to add more it defaults to a different day(sometimes with todays food added into yesterday's date).  

As a side note I'm super excited to see an open source calorie tracker out here. This app is simple, clean, and doesn't nag me to subscribe to their annual service while Im trying to input foods.

corngood:
Fixes #52

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -20,7 +20,7 @@
 var diary = {
 
   category:"0", //Category index
-  date: undefined,
+  date: app.getDateAtMidnight(),
   consumption:{}, //Nutrition consumed for current diary date
 
   setCategory : function(index)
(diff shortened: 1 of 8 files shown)


corngood:
@davidhealey @Mitlik I'd appreciate some code review on this.

I've added a _version field to the export json to indicate the corresponding DB version.  The upgrade logic (for records) is now done on db upgrade and json import.

- compatibility of the db and export data with older versions is untested and probably a bad idea

- use of local time is very limited now:
  - `app.getDateAtMidnight()` to get the UTC time of midnight for the current day (default diary day, etc)
  - food item `dateTime`, which I believe is just used for ordering?

- import completion handling is now done on the transaction instead of request

davidhealey:
Thanks, I'll check it out properly as soon as I have some time.

PR #95, opened 2018-11-13
remove date filter from food list

corngood:
This was incorrect after the UTC change because food item dates could be
more than one day ahead of UTC midnight.

Passing null will return all records, which I believe is the intent.

This fixes a bug introduced in b01f0be0d7e8d91591dd548c21e6ab4dee539645 where a food item disappears from the food list after being added to the diary, because the new timestamp is more than one day ahead of UTC midnight.


davidhealey:
That's great, I didn't think to use null!

PR #96, opened 2018-11-14
fix some incorrect promise chains

### LINKED ISSUE #75: Homescreen display issue
Finally 2.3.5 has hit F-Droid so I am able to confirm the following was not just a bug with the older version I had access to.

* LineageOS: 14.1 (Android 7.1.2)
* Waistline 2.3.3 - 2.3.5

**Issue:**
Changing `homescreen` in settings does not work as expected.   It results in a blank version of the chosen screen

**What happens:**
Go to `settings` and change homescreen to diary.   Back out of waistline and reopen it from Android.   The correct screen is selected but the contents are black.  To display the contents you need to click the hamburger in top left of waistline and re-select your chosen homescreen.

This appears to only be if you choose `statistics`,`diary`, `food list`, `recipes` and `goals`.

It does not happen if you select, `settings` or `user guide` as your homescreen.

**What should happen:**
Go to `settings` and change homescreen to diary.   Back out of waistline and reopen it from Android.   The correct screen is displayed correctly.

corngood:
While investigating #75, I found `.then(resolve())`, which was causing the first page to load before the database was initialized.  I searched for other suspicious uses of `then`, and found a couple more.

Fixes #75

### DIFF
--- a/www/activities/meals/js/meals.js
+++ b/www/activities/meals/js/meals.js
@@ -333,7 +333,7 @@ $(document).on("show", "#edit-meal", function(){
   {
     $("#edit-meal #title").html(app.strings["recipes"]["edit-recipe"]["title2"]);
     meals.fillEditForm(this.data) //Populate edit screen with data
-    .then(meals.validateEditForm());
+    .then(() => meals.validateEditForm());
   }
 });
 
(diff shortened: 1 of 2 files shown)


davidhealey:
Wonderful, thanks

PR #98, opened 2018-11-15
Food list search fixes

corngood:
I have a lot of problems with the food list search box if I rapidly change the search text.  Sometimes duplicate items will appear, sometimes the app will become unresponsive.  If you have a big list of food items on an android device, try rapidly entering and removing a single char from the search box, and you'll probably see what I mean.

I think these are mostly caused by the async update when you clear the search box.  Rather than making async searches work reliably, which would involve queueing or cancelling them somehow, I just took out the DB update.  If you can think of a reason it needs to be there, let me know, but filtering never changes the master list of food items, so I don't see why it would be required.

It seems much more stable to me now.

### DIFF
--- a/www/activities/food-list/js/food-list.js
+++ b/www/activities/food-list/js/food-list.js
@@ -600,7 +600,7 @@ $(document).on("init", "#food-list-page", function(e){
   });
 });
 
-$(document).on("keyup", "#food-list-page #filter", function(e){
+$(document).on("input", "#food-list-page #filter", function(e){
 
   $("#food-list-page ons-toolbar-button#submit").hide(); //Hide submit button until items are checked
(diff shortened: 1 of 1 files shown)


davidhealey:
Yeah I'm not sure what I was thinking :) I think at some point it would also be good to make the list lazy load as well but that's not a priority for now

corngood:
Yeah, I guess I'll have to see what it's like when I have a year worth of food items instead of a month.  Right now performance is totally fine for me without lazy loading.

Cheers

corngood:
I just found out why the database update was there.  When you clear after an OFF search it's needed to get your local food items back.

I'll make sure to fix it tomorrow.

davidhealey:
Oh that makes sense. If the database update is happening on every search though maybe we could make it so it only happens when the search box is empty.

Issue #66, opened 2018-09-22
Waistline 2.3.4: minSdkVersion 16 cannot be smaller than version 19 declared in library [:CordovaLib]


relan:
I'm trying to build Waistline 2.3.4 on a Debian 8.11 system with Gradle 4.1. It fails with

```
Execution failed for task ':app:processReleaseManifest'.
> Manifest merger failed : uses-sdk:minSdkVersion 16 cannot be smaller than version 19 declared in library [:CordovaLib] /home/fdroid/fdroiddata/build/com.waist.line/platforms/android/CordovaLib/build/intermediates/manifests/full/release/AndroidManifest.xml as the library might be using APIs not available in 16
  	Suggestion: use a compatible library with a minSdk of at most 16,
  		or increase this project's minSdk version to at least 19,
  		or use tools:overrideLibrary="org.apache.cordova" to force usage (may lead to runtime failures)
```

Steps to reproduce:

```
git checkout 2.3.4
cd platforms/android/app
gradle assembleRelease
```

Any clues?

P.S. Waistline 2.3.3 builds just fine on the same system.

davidhealey:
The SDK version required was changed for 2.3.4 (I'm not sure why, I think it was a requirement of one of the plugins that I updated).

davidhealey:
Are you able to build 2.3.3?

relan:
Yes, 2.3.3 builds just fine on the same system.

davidhealey:
Hi,

I'm just looking into this issue now. If you open AndroidManifest.xml file in /platforms/android/app/src/main/ and change this line <uses-sdk android:minSdkVersion="16" android:targetSdkVersion="27" /> to this <uses-sdk android:minSdkVersion="19" android:targetSdkVersion="27" /> Are you able to build?

davidhealey:
Actually I've just made this change and pushed it to the repo, see if you can build the master branch now, if you can I'll push a new release

relan:
Hi,

I'm on vacation now, away from my PC. Will be able to test only in two weeks.

davidhealey:
No problem

relan:
2.3.5 and 2.3.7 were built successfully, so this issue is solved. Thanks!

PR #150, opened 2019-05-02
Rename fibre in goals.html to fix statistics bug

AlexGitReid:
I encountered a bug on the statistics screen when swiping through the macronutrients in the version of the app on the play store.

I expected to see all of the macros alongside my goals and current daily amounts, but the Fiber goal was fixed at 24, even though I set it to a different value, and my daily Fiber was being assigned to another macro. One of the statistics values was "undefined", as well, which contained my set goal for Fiber.

After looking at the code, I believe goals.html was the culprit. On this page, id and data-localize were set to "fibre", but the localization files (such as locale-en.json) use the string "fiber" for the id.

Changing goals.html and rebuilding/loading to my phone resulted in the statistics screen displaying correctly.

I did not take any screenshots of the behavior before or after, let me know if you want them or would like more details.

Side note: When making my changes, a bunch of files in the platform folder were modified by Cordova. I committed them, but I also created a branch that removes the build artifacts with `.gitignore`. Would you like me to post that as well?

Based on (https://github.com/davidhealey/waistline/issues/66), it looks like those build artifacts may be required by F-Droid? I won't touch them for now as part of this PR.


davidhealey:
Hi, thanks for the contribution. I'm currently doing quite a large refactor/rewrite on an unpushed branch and I'll be sure to include your fixes in the next update too. Yes some of the platforms stuff is necessary, if you're using the gitignore from the master branch then it should be correct.

AlexGitReid:
Thank you, David!
If you are ever looking for testers for your refactor, let me know! I am a fan of the app and would enjoy helping out.

davidhealey:
If you get notification you'll see when I push the new branch and you can try it out

PR #236, opened 2021-05-04
Add settings to export

jfaltis:
This PR also includes the users settings when exporting and importing (if present)


davidhealey:
That's a good idea!

dagmarkrefting:
amazing app, but I found it difficult to import my old database when changing my device. 
I could not find a documentation, so I had to find the export in my old device, copy it to the new device (where the path changed from android/data to android/media, which I found out by the error message) and then import it.
export and import settings for file selection would be great - or probably a short description in the readme how to do it :-)

PR #241, opened 2021-05-20
Fix search for meals and recipes

jfaltis:
I noticed that you can search through foods but not meals and recipes. This PR fixes this.


davidhealey:
Thanks, looks like I had some typos from when I was refactoring.

PR #243, opened 2021-05-30
Fix misassigning log object to weight in db-handler

asdomingues:
A logs object from the old schema was beign assigned to weight property
in the new schema, yielding unexpected behaviour even though the import
completed successfully.

Example of log object (waistline_export.json -> log[0]):
```json
{
    "dateTime": "2021-05-03T00:00:00.000Z",
    "nutrition": {
        "calories": 2461.0960000000005
    },
    "weight": "72.5",
    "goals": {
        "weight": {
            "target": "67.95",
            "weekly": "0.629",
            "gain": false
        }
    }
}
```

Actual migration result (waistline_export.json -> diary[0].stats):
```json
{
        "weight": {
            "dateTime": "2021-05-03T00:00:00.000Z",
            "goals": {
                "weight": {
                    "gain": false,
                    "target": "67.95",
                    "weekly": "0.629"
                }
            },
            "nutrition": {
                "calories": 2461.0960000000005
            },
            "weight": "72.5"
        }
}

```

Expected migration result:
```json
{
        "weight": "72.5"
}

```

### DIFF
--- a/www/assets/js/db-handler.js
+++ b/www/assets/js/db-handler.js
@@ -223,11 +223,11 @@ var dbHandler = {
 
             store.openCursor().onsuccess = function(event) {
               let cursor = event.target.result;
-              if (cursor) {
+              if (cursor && cursor.value && cursor.value.weight) {
                 let value = cursor.value;
                 let dateTime = value.dateTime.toDateString();
 
-                weights[dateTime] = value;
+                weights[dateTime] = value.weight;
                 cursor.continue();
               } else {
                 resolve();


www/assets/js/db-handler.js
```diff
@@ -223,11 +223,11 @@ var dbHandler = {
 
             store.openCursor().onsuccess = function(event) {
               let cursor = event.target.result;
-              if (cursor) {
+              if (cursor && cursor.value && cursor.value.weight) {
```
asdomingues, line 226: Hi @davidhealey, the `weight` property appears to be always set, but I'm not sure if that property holds. ALso, is there an easy way to test this repo?

davidhealey: There shouldn't be a weight goal at all in the new version, since it isn't used for anything. The weight log is now part of the diary and exists in the stats object.

> Also, is there an easy way to test this repo?

That depends on what you mean by easy. I do 99% of my development in Chromium and only use an actual Android device for testing and debugging.

asdomingues: Yes, I wasn't clear. I meant I don't know if there is a need for  the two checks I added... What do you think?
I'll try to run chromium with the inspection feature then.

davidhealey: Oh I see, I can't remember exactly what values are there so I'll need to test it and monitor the output to know for sure.

asdomingues: I don't know which value is there since the beginning (if you never set a weight). But if you have set it at least once, I think this value is copied to the next day until a new one is inserted. But I'll let you know if I'm able to test it

davidhealey: I looked into this today and fixed a few things with the importing/upgrading so we should be good to go now

PR #244, opened 2021-05-31
Update carbs, protein and sugar to new names in db-handler

asdomingues:
Those properties were renamed recently and `foodList` is being migrated with
the old names yet, statistics use foodList's nutrition facts to
aggregate data based on diary reference to foodList's id for each item.
Thus, statistics screen doesn't show carbohydrates, proteins and sugars.

Example of old `item.nutrition`:
```json
"nutrition": {
    "calories": 137,
    "carbs": 25.95,
    "fat": 1.5,
    "fiber": 1.5,
    "protein": 4.4,
    "salt": 0,
    "saturated-fat": 0.32,
    "sodium": 304000,
    "sugar": 0.12
}
```

Example of `item.nutrition` migrated from old version to new one:
```json
{
    "calories": 137,
    "carbs": 25.95,
    "fat": 1.5,
    "fiber": 1.5,
    "protein": 4.4,
    "salt": 0,
    "saturated-fat": 0.32,
    "sodium": 304000,
    "sugar": 0.12
}
```

Example of `item.nutrition` created directly at the new version:
```json
{
    "calories": 137,
    "carbohydrates": 25.95,
    "fat": 1.5,
    "fiber": 1.5,
    "proteins": 4.4,
    "saturated-fat": 0.32,
    "sodium": 304,
    "sugars": 0.12
}
```

### DIFF
--- a/www/assets/js/db-handler.js
+++ b/www/assets/js/db-handler.js
@@ -309,6 +309,11 @@ var dbHandler = {
         await new Promise(function(resolve, reject) {
           store = transaction.objectStore('foodList');
           let foods = [];
+          const renamedProperties = {
+            "carbs": "carbohydrates",
+            "protein": "proteins",
+            "sugar": "sugars"
+          };
 
           store.openCursor().onsuccess = function(event) {
 
@@ -316,6 +321,15 @@ var dbHandler = {
 
             if (cursor) {
               let item = cursor.value;
+
+              if (item.nutrition !== undefined) {
+                Object.keys(renamedProperties).forEach(oldProperty => {
+                  if (item.nutrition[oldProperty]) {
+                    item.nutrition[renamedProperties[oldProperty]] = item.nutrition[oldProperty];
+                    delete item.nutrition[oldProperty];
+                  }
+                })
+              }
 
               if (item.portion !== undefined) {
                 item.portion = parseInt(cursor.value.portion);


www/assets/js/db-handler.js
```diff
@@ -317,6 +322,15 @@ var dbHandler = {
             if (cursor) {
               let item = cursor.value;
 
+              if (item.nutrition !== undefined) {
+                Object.keys(renamedProperties).forEach(oldProperty => {
+                  if (item.nutrition[oldProperty]) {
```
asdomingues, line 327: Old ones could be `null`, hence I did not check explicitly by `undefined`

davidhealey:
Looks like a good solution, what else needs doing before merge?

asdomingues:
> Looks like a good solution, what else needs doing before merge?

I didn't test it yet :slightly_frowning_face:

asdomingues:
Hey @davidhealey, I can't manage to install this app version even after accepting to install it with a different signature (checked by Play Protect). Could you create a debug app for this one in order for me to check if it works?

davidhealey:
Here you go
[app-debug.zip](https://github.com/davidhealey/waistline/files/6715569/app-debug.zip)

davidhealey:
Aww I totally forgot about this pull request! Your method is more elegant than what I did.

asdomingues:
Oh, really? Sorry! I just wasn't able to test even with your app version. There is an error installing this version. As I didn't have the time to walk through the log cat and I thought it was already solved, I closed the issue.

Issue #285, opened 2021-08-03
Use nutrition facts for the prepared product of OFF entries when available


EmilJunker:
Open Food Facts differentiates between "nutrition facts for the product as sold" and "nutrition facts for the prepared product". Most products only specify nutrients "as sold", but for certain items (for example pudding powder), it's much more interesting to get the "prepared product" nutrients as I don't really care how much fat/salt/etc. is in the powder itself. After all, I'm not going to eat the powder but the prepared pudding...

In fact, here is an example of an OFF entry that **only** specifies nutrition facts for the "prepared product" and has **no information** about the product "as sold":
OFF website: https://world.openfoodfacts.org/product/8710604724739/spaghetti-alla-carbonara-knorr
OFF API link: https://world.openfoodfacts.org/api/v0/product/8710604724739.json

When I try to scan the barcode of the above product with Waistline, it says "No matching results". Apparently, you are only checking for the nutrients "as sold" at the moment; and if a product doesn't have them, you ignore it.

I think if an OFF product comes with "nutrition facts for the prepared product", Waistline should use them.

davidhealey:
Thanks, I'll add it to the list, definitely shouldn't return no matching results.

davidhealey:
https://github.com/davidhealey/waistline/pull/286

PR #286, opened 2021-08-04
Use OFF nutriments for the prepared product when nutriments as sold are not available

EmilJunker:
I went ahead and implemented the changes proposed in #285 myself.

### DIFF
--- a/www/activities/foodlist/js/open-food-facts.js
+++ b/www/activities/foodlist/js/open-food-facts.js
@@ -80,12 +80,12 @@ app.OpenFoodFacts = {
   },
 
   parseItem: function(item) {
-    const nutriments = app.nutriments; //Array of OFF nutriment names
+    const nutriments = app.nutriments; // Array of OFF nutriment names
     let result = {
       "nutrition": {}
     };
 
-    //Search for all keys containing 'item_name' to include local item names
+    // Search for all keys containing 'item_name' to include local item names
     for (let k in item) {
       if (k.includes("product_name") && item[k].length > 1) {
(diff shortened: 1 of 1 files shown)


davidhealey:
Thanks, I'll review it at the weekend

njmdietrich:
If I may butt in, I don't think the prepared product should be the default values used. Personally, I like to weigh things out before preparing them, and I would imagine most people planning their meals by calories/nutrients (rather than just tracking them) do the same. So I think if available, nutrients as sold are preferable, perhaps with a function to choose which information to use.

EmilJunker:
@njmdietrich I think you misunderstand the meaning of "prepared product". It only affects products that are not supposed to be eaten as sold anyway. The reason this bug hadn't been noticed by anyone until now is because the vast majority of products don't even specify nutrition info for the prepared product. It only affects things like pudding powder or dried products to be rehydrated.

If an item specifies nutrition info for the prepared product, then it makes perfect sense to use it.

EmilJunker:
Now that I think about it, I believe some mueslis actually specify the nutrition info for both the product as sold and the prepared product (the latter including the milk). In cases like this, it would indeed be preferable to use the nutrients as sold over the ones for the prepared product.

Alright, I've now changed it so nutriments for the prepared product are only used when the nutrients as sold are not available.

njmdietrich:
Yes, I also know some frozen foods etc have nutriments for the prepared product, i.e. when cooked, in which case it also makes sense to have the "as sold" info for weighing things out before cooking.

Issue #183, opened 2019-09-12
Option for switching off barcode scanning sound


mahlzahn:
Loud beep is unexpected for such an app and might be annoying …

davidhealey:
Note to self - https://stackoverflow.com/questions/40984987/how-to-disable-beep-on-phonegap-plugin-barcodescanner-meteor-and-cordova

davidhealey:
https://github.com/davidhealey/waistline/commit/64ab332ea6fa50b95d4bc34ad7f833d01bc54e3d
https://github.com/davidhealey/waistline/commit/c5c9fb8e64df093eb95d5a66a8203c4712b6dbe5

PR #287, opened 2021-08-04
Pass parameters to barcode scanner plugin

EmilJunker:
This PR adds two parameters to the barcodeScanner function call. The loud beep sound mentioned in #183 has bugged me for quite a while, so I added a parameter to disable it. While I was at it, I also added a parameter to show a toggle button for turning the phone's torch light on and off. This could be useful for scanning a barcode when there is insufficient ambient lighting.

I successfully tested these features on my own Android phone to make sure that they work.

I guess one could also add a setting to enable/disable these features, but I don't think that's necessary. I can't really imagine anyone who would prefer to hear a loud sound every time they scan something. And the torch toggle button is only displayed in the corner as an option, no one is forced to use it.

### DIFF
--- a/www/activities/foodlist/js/foodlist.js
+++ b/www/activities/foodlist/js/foodlist.js
@@ -357,6 +357,13 @@ app.Foodlist = {
         } else {
           resolve(undefined);
         }
+      },
+      async (error) => {
+        resolve(undefined);
+      },
+      {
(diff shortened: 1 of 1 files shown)


davidhealey:
This is a good one. I think I will add an option for the beep sound because I like the sound :)

PR #339, opened 2021-09-09
Add adaptive icons and landscape splashscreen

### LINKED ISSUE #313: Add Android Adaptive icon
Android 8.0 introduced the concept of Adaptive icons. The idea is to split the app icon into a background and a foreground instead of just providing a single static app icon with a predefined shape. This way, the Android launcher can apply a consistent shape to all icons on the home screen for a more uniform look and feel. Some launchers can even animate the icons using a parallax effect where the background and foreground are sort-of "shaken" independently from each other. For more information see [the docs](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive).

Cordova has built-in support for Adaptive icons, see [here](https://cordova.apache.org/docs/en/latest/config_ref/images.html#android). All you need to do is provide the `background` and `foreground` images in addition to the `src` image in the `config.xml.` file.

The current Waistline app icon looks like this:

xxxhdpi

I would suggest to split this into a background and foreground as follows (This is just a sketch. I'm not a designer, as you can probably tell 😅)

The background consists of a solid red background with the scale printed on it:

bg

The foreground is mostly transparent with the letter W in white and the pointer in red:

fg

Put the background and foreground together and you get the following. If the icon was animated (i.e. parallax effect) this should hopefully give an interesting result with the pointer moving along the scale.

icon

With a round icon shape applied, it would look like this (the black portion gets removed):

circle

With a "squircle" icon shape applied, it would look something like this (again, the black portion gets removed):

squircle

This way, the app icon could adapt to different icon shapes (whichever shape the launcher is using) and thus would fit in better with other Adaptive icons on the home screen.

### LINKED ISSUE #331: Missing landscape orientation splashcreen

EmilJunker:
This adds adaptive icons and a horizontal orientation splashscreen.

Here's what the adaptive icon looks like with different shapes applied:

Circle:
Screenshot_20210909-182950

Squircle:
Screenshot_20210909-183029

Square:
Screenshot_20210909-183124

Teardrop:
Screenshot_20210909-183235

Rounded: (also how the fallback icon looks on older phones)
Screenshot_20210909-183156

The new landscape splashscreen looks a bit distorted on my phone:
Screenshot_20210909-183409

But the portrait splashscreen always looked a bit distorted too actually, so it's no big deal:
Screenshot_20210909-183310

Closes #313, closes #331.

By the way, here is my edited SVG in case it is needed: [WaistlineIconSVG.zip](https://github.com/davidhealey/waistline/files/7138484/WaistlineIconSVG.zip)


EmilJunker:
I also added default density splashscreen images now. This should fix the `MissingDefaultResource` error some people were getting and that I believe was also in the fdroid build logs.

And I was able to fix the distorted splashcreen. Looks much better now in both landscape and portrait mode:

Screenshot_20210910-092957

Screenshot_20210910-092930

davidhealey:
Why are we losing these plugins?

```
<plugin name="cordova-plugin-camera" spec="~4.0.3" />
    <plugin name="cordova-plugin-file" spec="^6.0.1" />
    <plugin name="cordova-plugin-network-information" spec="~2.0.1" />
    <plugin name="cordova-plugin-whitelist" spec="^1.3.3" />
    <plugin name="cordova-plugin-splashscreen" spec="^5.0.2" />
    <plugin name="phonegap-plugin-barcodescanner" spec="^8.0.0">
        <variable name="ANDROID_SUPPORT_V4_VERSION" value="23.+" />
    </plugin>
```

EmilJunker:
I removed the plugin references from the config.xml because they are redundant. All plugins are declared in package.json (which apparently has higher priority than config.xml).

I now resolved the conflict, so this PR can also be merged.

PR #348, opened 2021-09-16
Fix new nutriments missing after upgrade/import

EmilJunker:
This fixes a bug where the nutriments I added yesterday are missing when upgrading or importing from the previous version. Sorry I missed this.


davidhealey:
Thanks.

I haven't studied it in detail yet but I think we can just loop through the main nutriments array and if one is missing from the Settings then we add it. That way we don't have to write the same thing in two places. I'll look at this more at the weekend.

davidhealey:
Oh I see you improved the nutriments thing. Looks like it's good to go.

EmilJunker:
Yes I did. I also added a settings schema version to prevent it from checking the full list of nutriments every time the app is launched.

I'd say it is time to see if the fdroid build works now.

PR #355, opened 2021-09-28
Bugfixes

### LINKED ISSUE #330: Food names with quotations not rendering properly

### LINKED ISSUE #353: Notes still showing in Diary editor even when disabled

EmilJunker:
This fixes #353 and fixes #330.


davidhealey:
What changed to fix the notes visibility? I'm looking at the diff but I must be missing the difference...

EmilJunker:
The difference is that the visibility setting check now happens outside the if-statement. Before it wouldn't be checked when coming from the diary. Now it is always checked.

PR #363, opened 2021-10-02
Add support for right-to-left writing direction

EmilJunker:
This is an attempt to add proper RTL (right to left) language support to the app. I added a setting for switching between LTR and RTL. It's not perfect and there are probably some rough edges here and there, but it's a start.

Here are some screenshots (in Hebrew as it's currently the only supported language for which it makes sense).

Diary page:

localhost_1

Diary settings:

localhost_2

Food editor:

localhost_3

@yarons I feel like you're the expert here. Does this look usable? Is this an improvement over the current version?


yarons:
Hey @EmilJunker , great work, thank you!

I just have some small comments :)
Hebrew acts just like English when referring to proportions, as you can see the Calories, Fat, Carbs all appear as 250 / 0 instead of 0 / 250 (These should be kept LTR in that case).

Can you send me the APK so I can test it further on my device?

Regardless of this effort, in Hebrew we're using a localized units so kcal should be in Hebrew (קק״ל) and g should be ג׳.

Thank you!

EmilJunker:
@yarons Here's an updated APK (since it's built by me it is unsigned and you probably need to uninstall the current verion before installing mine)

[app-debug.zip](https://github.com/davidhealey/waistline/files/7272933/app-debug.zip)

davidhealey:
> @yarons you probably need to uninstall the current version before installing mine)

And back up your data first

yarons:
Sorry for mixing other issues but there are few glitches.

IMAGE 2021-10-09 23:12:40
There are some untranslated strings here.

IMAGE 2021-10-09 23:23:09
Energy is still in English.

IMAGE 2021-10-09 23:23:42
These should be aligned right and the radio button should be on the left.

IMAGE 2021-10-09 23:24:35
This is what happens when my device is configured for Dark mode and the app doesn't.

BTW Changing the Dark mode toggle didn't change much, it was dark before selection.

Overall you did a very good job, thank you so much.

EmilJunker:
> There are some untranslated strings here.

> Energy is still in English.

I'll add the missing strings. Although I don't think we need to localize all the country and language names in the dropdown menus. It's fine if they stay in English.

> These should be aligned right and the radio button should be on the left.

This is not possible according to Stackoverflow: https://stackoverflow.com/questions/28146034/how-to-make-text-align-right-in-selectbox-list-in-android-chrome. If anyone knows a way to do it, please tell me.

> This is what happens when my device is configured for Dark mode and the app doesn't.

> BTW Changing the Dark mode toggle didn't change much, it was dark before selection.

That's weird, On my phone, the device dark mode has no effect on the app. And the text should definitely become white when you enable the in app dark mode. Is this related to the RTL layout? What happens when you switch to a different language that doesn't use RTL?

PR #364, opened 2021-10-02
Fix missing context menus

### LINKED ISSUE #251: Can't copy/paste into text fields
I'm using the latest build from #217. When typing in any text field, I can't long press or use my keyboard actions to copy or paste text. I'm on Android 10, pretty close to stock, on the OnePlus Nord N10 5G.

EmilJunker:
Removing the following lines from index.js fixes #251 (missing copy/paste option) on my phone.

```js
//Prevent chrome displaying context menu on long click
window.addEventListener("contextmenu", (e) => {
  e.preventDefault();
});
```

### DIFF
--- a/www/index.js
+++ b/www/index.js
@@ -409,6 +409,13 @@ document.addEventListener('deviceready', async function() {
   }, 2000);
 }, false);
 
+// Prevent chrome displaying context menu on long click
+window.addEventListener("contextmenu", (e) => {
+  let target = e.target.nodeName.toLowerCase();
+  if (target !== "input" && target !== "textarea")
+    e.preventDefault();
+});
+
 // Android back button 
 document.addEventListener("backbutton", (e) => {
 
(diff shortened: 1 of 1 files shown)


davidhealey:
Won't this create issues with long press on the quick-add button for example?

EmilJunker:
No, long press on the quick add button still works. But my phone no longer vibrates when I do that (it did before). I'll try to get that fixed...

davidhealey:
I'll have to test on mine too. I'm sure I disabled that behaviour for a reason.

EmilJunker:
I've now changed it to

```js
window.addEventListener("contextmenu", (e) => {
  let target = e.target.nodeName.toLowerCase();
  if (target !== "input" && target !== "textarea")
    e.preventDefault();
});
```

This should serve the original purpose (whatever that was) while making sure that context menus work in text input fields. This also fixes the broken vibration.

davidhealey:
That looks like a good solution

Issue #368, opened 2021-10-04
Anchor the calorie meter to the screen in the diary instead of scrolling it away as you enter food


mrhebisan:
In the older version of the app and in other apps I have used, the daily calorie count stays visible in the diary no matter how many food items were entered.  Can this be restored as an option?

It now scrolls away since it is now at the top of the scrollable view instead of being anchored. This is a degradation in function from my perspective.

 Keeping constant eyes on that calorie goal is very helpful while editing later day meals, or preplanning a whole day in advance.

  If I pre-enter the food before I eat it I have to constantly scroll up and down the screen to make sure I'm not overeating with every new entry.

davidhealey:
Yes I think this is a good idea it annoys me too. I want to make some other changes to the calorie display anyway. Waistline's number 1 purpose is to be a calorie counter so I want to give the calorie consumption priority in the diary display. I don't like the current setup and I have a few plans for improvement.

EmilJunker:
I've experimented a bit and came up with a layout where the date switcher stays fixed at the top and the daily energy/nutrient counts stay fixed at the bottom. The content in between is scrollable.

![Screenshot 2021-10-04 143147](https://user-images.githubusercontent.com/19289477/135852451-60f4dcba-671a-4a83-bf7d-ccdf29150294.png)

If it's alright I can create a pull request so this can be included in the next release already.

PR #369, opened 2021-10-04
Improve diary layout

EmilJunker:
This makes the date switcher stay fixed at the top and the daily energy/nutrient counts stay fixed at the bottom. The content in between is scrollable.

See #368.

### DIFF
--- a/www/activities/diary/views/diary.css
+++ b/www/activities/diary/views/diary.css
@@ -18,6 +18,6 @@
 */
 
 /* Nutrition/Goals display */
-.page[data-name="diary"] #diary-nutrition {margin-top: 1em; margin-bottom: 1em;}
-.page[data-name="diary"] #diary-nutrition .col {text-align: center;}
+.page[data-name="diary"] #diary-nutrition .row {justify-content: space-around;}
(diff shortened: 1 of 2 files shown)


davidhealey:
Looking this in the browser, emulating a Galaxy S5, there is an issue with the positioning of the nutriment consumed/goals. As I said elsewhere I have plans to change this completely in the future, but for now I think we should have this working.

Peek 2021-10-09 13-58

EmilJunker:
> Looking this in the browser, emulating a Galaxy S5

I can't replicate this on my machine. When I emulate a Galaxy S5, it only displays three stats per page and it looks alright.

Can you inspect and play around with the flexbox properties a bit? I think I changed it from `space-between` to `space-around`, although I can't imagine that's the reason for the issue 🤔

EmilJunker:
> Peek 2021-10-09 13-58

The issue is not completely fixed. Now I'm seeing it too on my phone. It's very weird though, sometimes it's there and sometimes it isn't. Also, rotating my phone into landscape view and then back seems to fix it.

Unfortunately I still can't replicate this in the emulator on my PC which would make it easier to debug.

EmilJunker:
> You can use cordova run android -- --livereload to debug in realtime on your phone.

Thanks for the tip. That doesn't seem to work from within a Docker container though, so I can't use it. Anyway, I think I've fixed it for good now in #374.

PR #375, opened 2021-10-09
Fix OFF search and barcode scan

EmilJunker:
Fixes a major bug that prevents searching for and importing OFF items.

Also updates package-lock.json to close 2 high severity vulnerabilities in some node packages.


davidhealey:
Should I push a new release with this fix?

PR #391, opened 2021-10-31
Show average goal in statistics

EmilJunker:
At the moment, the goal line on the statistics page always shows the goal for the current day. This is fine, of course, if you use the same goal every day. But it isn't that useful when you have a different goal for each day or when the "auto adjust" setting is enabled. In that case, it would make more sense to show the average of your goals because then you can more easily compare your average consumption (red line) to your average goal (green line).

This PR changes the goal line on the statistics page to show the average goal instead of just the goal for the current day.

### DIFF
--- a/www/activities/diary/js/group.js
+++ b/www/activities/diary/js/group.js
@@ -116,7 +116,7 @@ app.Group = {
 
     //Energy 
     const energyUnit = app.Settings.get("units", "energy");
-    const energyName = Object.keys(app.nutrimentUnits).find(key => app.nutrimentUnits[key] === energyUnit);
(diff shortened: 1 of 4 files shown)


davidhealey:
I think this should have a switch in the settings page

EmilJunker:
> I think this should have a switch in the settings page

I initially thought that too, but I've concluded that it's not really necessary. After all, the purpose of the statistics page is to display information about a longer period of time (the smallest range you can pick for the diagram is "1 week"). Is there any scenario where someone would go to the statistics page to see their goal for the current day? Isn't that what the diary is for?

davidhealey:
Ah that's a very good point. You've convinced me :)

PR #392, opened 2021-11-03
Food editor improvements

EmilJunker:
This PR includes some improvements for the food editor and fixes a bug in the OFF import.

In the food editor, the food image now has a `max-width` and a `max-height` instead of an explicit `width`. This looks much better, in my opinion. Here are some examples before (left) and after (right):

arrabbiata1 arrabbiata2

A relatively wide image before and after:

milka1 milka2

A relatively tall image before and after:

soja1 soja2


davidhealey:
I like it!

Issue #349, opened 2021-09-17
Improve food editor input


davidhealey:
Especially in the diary page when you just want to make a quick change to the serving or quantity, it's clumsy to be in a text box having to manually delete values one character at a time. I think it would be better to open a modal input window with an empty input box and default to numeric input.

ZMcK:
Most other food tracking apps I've seen use a wheel/picker for the primary adjustment and a numeric input as backup. I've used Lose It! for several months now and it works very well. I see Framework7 has a picker tool. If you're interested in using it, I'd work on implementing it.

davidhealey:
Yeah a picker sounds like a good idea. I'll explore some options this weekend and let you know my thoughts. Whatever we go with is needs to be used for editing serving size, number of servings, and possibly nutriment values, although maybe nutriments can remain as is.

davidhealey:
I think for now I'm going to stick with the input boxes. But I'm going to change their behaviour slightly so that when you click them it automatically highlight the text in the box so you can type straight away to replace it. I'm also going to change the styling of the input boxes to make them more clearly defined.

mrhebisan:
Given that I try to keep serving sizes in grams, I'm glad to see you not choose a spinner.  I'm definitely interested in the field auto highlighting.  Maybe even just flat out auto selected when the editor is open?

As is, the margin on the containing editor is too close to the edge of the screen and I often cannot drag the selector around to the far edge.  Its very frustrating.

ZMcK:
Just putting it out there that I'd implement the picker to have an auto converter, so any food that pops up in oz, lbs, etc can be auto converted to grams. FYI, only pushing for this because I use one on the aforementioned app literally all day everyday and it works super well.

davidhealey:
> any food that pops up in oz, lbs,

How would it deal with none standard units. Like 1 biscuit or 1 slice?

Issue #310, opened 2021-08-28
Quantity popup


davidhealey:
When adding a single item to the diary/meal/recipe prompt for the quantity. Also add a setting to disable this.

EmilJunker:
Maybe a bit more difficult to implement, but when adding multiple items at once, the app could prompt for the quantity of each one by one. Just an idea...

davidhealey:
Shouldn't be too difficult to implement, just a single command vs a loop I think. I'll add an option to toggle that though because some people might not want it.

h-h-h-h:
Could this be extended with a calories textbox? Not only would this allow you to quickly make a rough estimate what amount of the food you're allowed to consume ("Would 2 cookies be okay? Would 3 cookies be okay?"), but you could also enter a calorie value and, e.g., get the weight you can eat. See also what I wrote [here](https://github.com/davidhealey/waistline/issues/337#issuecomment-1016026648) about fraction-based food (this dialog would then adhere to the default linking defined for the food).

If this dialog would add calories, it so neatly would contain all the centrally important values.

PR #398, opened 2021-11-05
Diary popup prompt for quantity

EmilJunker:
This adds the quantity popup mentioned in #310. There is also a setting to enable it (it's disabled by default). There is probably still some room for improvement (like showing the item's brand or calories in the popup) so feel free to change/extend this implementation as you like.

This also includes a function that auto-selects the text in any input field with class `auto-select` when the user taps in it. Maybe this can be used for #349 as well.


davidhealey:
Looks good. I like the flatten functions!

PR #399, opened 2021-11-05
Fix and improve food thumbnails alignment

EmilJunker:
Improving the alignment of food thumbnails in the diary, especially for the RTL layout.

This is how it looked before:

before

And this is how it looks now:

after


yarons:
I need to populate my app with some data to catch these kind of glitches, thank you so much!

EmilJunker:
> What's going on with the labels at the bottom?

The Crowdin branch hasn't been merged yet, that's why some of the strings appear in English. This won't be the case when the next release is pushed (assuming the Crowdin branch is merged before that).

davidhealey:
Next release tomorrow I was thinking, unless any objections?

EmilJunker:
> Next release tomorrow

Sounds good to me 👍

Issue #380, opened 2021-10-21
[Feature] Make It Clear When Editing/Uploading To OpenFoodFacts & Link


seniorm0ment:
Whenever you edit entries, it's not clear whether you are uploading/editing those on OpenFoodFacts or only locally.
If only locally, then there is no clear way to upload your additions to OFF. If it is editing to OFF, then I suggest making that more clear to the user. 

Also sugesting adding a link to all OFF entries.

PR #402, opened 2021-11-13
Add button to re-download imported food item info

EmilJunker:
This adds a button in the food editor to re-download imported food item information. The button only shows up for items that have a barcode (i.e. items from OFF or USDA) and only when adding or editing them. Pressing it will replace the local values with the latest information available for the item (except for the notes which are preserved).

There are various use cases for this, but the most important one is when you scan the barcode of a product whose OFF entry is incomplete or outdated. In this case, you can simply tap on the barcode to be taken to the OFF website, complete the information there, and then switch back to Waistline and press the button to fetch the changes you just made. This eliminates the need to rescan the barcode and should address #380.


davidhealey:
Very nicely implemented

Issue #385, opened 2021-10-21
[Bug] F-Droid Version Missing Icon


seniorm0ment:
The icon for Waistline on F-Droid seems to be missing?

EmilJunker:
It worked! The icon and screenshots are now live:
https://f-droid.org/packages/com.waist.line/

The description still says that Waistline is in beta, though. I think to get rid of this, someone would need to make a pull request to remove the old description from the metadata file in the fdroiddata Gitlab repository:
https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/com.waist.line.yml

EmilJunker:
The f-droid listing has been updated again:
https://f-droid.org/packages/com.waist.line/

The "beta" label is gone and new links for translation and donations have been added.

I think this issue can be closed now.

davidhealey:
@EmilJunker Excellent, thank you

PR #404, opened 2021-11-13
Provide metadata for f-droid

EmilJunker:
This adds the metadata (icon, screenshots, description) for f-droid as discussed in #385. I also took this opportunity to create some new screenshots and update the Readme.

By the way, it looks like f-droid hasn't actually picked up the new release because the version code in `config.xml` wasn't updated. I'd suggest to tag a new release soon so f-droid can pick it up.


davidhealey:
Aww darn, I updated the version code and commited it but forgot to push. I'll make a new release tomorrow.

Issue #416, opened 2021-11-29
Recipe quick add with label


davidhealey:
Would be handy for adding one time items.

Tap + hold (like in Diary) should open a dialog to add a quick add item and should be able to enter a label if desired.

I think this label thing should be added to the diary quick add too (I think it was there in an earlier version and I removed it when changing the database schema.

PR #424, opened 2021-12-21
Diary quick add labels

EmilJunker:
Adds the option to set a description for quick add entries in the diary as suggested in #416.

localhost


davidhealey:
Is the description optional?

EmilJunker:
Yes, it is optional.

PR #445, opened 2022-01-22
Add setting for First Day of Week and Average Goal Base

### LINKED ISSUE #431: Carrying over calories to the next day
In Waistline v2, I always did this with my generic "Calories" entry, defined as 1 kcal. Depending on whether I ate more or less calories in a day, I entered a positive or negative number as the portion count; then, for the next day I entered the negated value.

Now, having upgraded to Waisline v3, negative portion counts aren't allowed anymore. Fortunately, the app at least doesn't crash because of my past negative diary entries.

- Would a quick, temporary fix be possible to allow negative numbers if, e.g., the entry is defined as having 1 kcal?!
- And then, would you please create a function to be able to carry over calories to the next day, no matter whether the daily goal is still green or it got red? I just want the average intake to adhere to my goal.

EmilJunker:
This adds a setting to choose the first day of the week and to specify how the average goals should be computed. Closes #431.

Untitled

### DIFF
--- a/www/activities/goals/js/goal-editor.js
+++ b/www/activities/goals/js/goal-editor.js
@@ -68,7 +68,6 @@ app.GoalEditor = {
     app.GoalEditor.el.autoAdjust = document.querySelector(".page[data-name='goal-editor'] #auto-adjust");
     app.GoalEditor.el.minimumGoal = document.querySelector(".page[data-name='goal-editor'] #minimum-goal");
     app.GoalEditor.el.percentGoal = document.querySelector(".page[data-name='goal-editor'] #percent-goal");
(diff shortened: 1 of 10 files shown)


h-h-h-h:
The setting on the page where you can set the calorie goal still speaks of a weekly goal, at least in German: "Automatische Anpassung für Wochenziel". This should probably be phrased more generically.

EmilJunker:
This is because the Crowdin branch with the translations hasn't been merged yet. Some locales aren't fully translated.

@huuhaa @nikets @utkabulka @asdomingues It would be great if you could complete the missing Finnish, Russian, Portuguese and Spanish translations over on Crowdin: https://crowdin.com/project/waistline

PR #449, opened 2022-01-22
Remember and restore last scroll position in diary

### LINKED ISSUE #260: Diary page scrolls up by itself after deleting an item
After deleting an item from the Diary, the page always scrolls up to the very top of the list by itself. This didn't happen in the previous version of the app and it's pretty annoying.

EmilJunker:
With the changes in this PR, the Diary now remembers the last scroll position before any add/edit/delete operation and restores it after the operation is complete. Closes #260.

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -21,6 +21,7 @@ app.Diary = {
 
   ready: false,
   calendar: undefined,
+  lastScrollPosition: 0,
   el: {},
   groups: [],
 
@@ -45,6 +46,8 @@ app.Diary = {
       app.Diary.groups = this.createMealGroups(); //Create meal groups
       this.render();
       app.Diary.ready = true;
+    } else {
+      app.Diary.lastScrollPosition = 0; // Reset last scroll position
     }
   },
 
@@ -178,6 +181,11 @@ app.Diary = {
     swiperWrapper.innerHTML = "";
 
     await app.Diary.renderNutritionCard(totalNutrition, new Date(app.Diary.date), swiper);
+
+    if (app.Diary.lastScrollPosition !== 0) {
+      $(".page-content").scrollTop(app.Diary.lastScrollPosition); // Restore last scroll position
+      app.Diary.lastScrollPosition = 0;
+    }
   },
 
   renderNutritionCard: async function(nutrition, date, swiper) {
@@ -449,6 +457,7 @@ app.Diary = {
     } else {
       // No more items to process -> write entry to DB and refresh page
       await dbHandler.put(entry, "diary");
+      app.Diary.lastScrollPosition = $(".page-content").scrollTop(); // Remember scroll position
       app.Diary.render();
     }
   },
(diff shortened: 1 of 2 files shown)


h-h-h-h:
I was just attending to my notes about this bug and researching in the issue tracker as you committed this. What a coincidence.

davidhealey:
I think this is a good way to fix it, very nice!

h-h-h-h:
I say this from the perspective of only using a single diary category. When I'm scrolled down, tap `+` on the bottom of the diary and an item is added, the item and the `+` button I need again right away are always scrolled out of view. If you'd establish saving [`Number.MAX_VALUE`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE), [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) or [`Number.POSITIVE_INFINITY`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/POSITIVE_INFINITY) in the meaning of "scrolled all the way to the bottom" and restore it like this after adding an item, the behavior would be better in my use case.

However, what about rarer cases (at least for me) where more items are added in one go than fit on the page, or when using the `+` button not from the last category? Would it generally be more desirable to put the first new item on top of the page or to scroll down by the height of the newly added items? Is that possible?

EmilJunker:
> the item and the `+` button I need again right away are always scrolled out of view

Are you aware that you can add multiple items in one go by searching, selecting items, entering a new search query, selecting more items, and so on... The app remembers all selected items from previous searches until you tap the `✔` button.

> Would it generally be more desirable to put the first new item on top of the page or to scroll down by the height of the newly added items? Is that possible?

I guess it's technically possible, but I don't know how to implement it in a safe way. I think this would be quite difficult to get right.

h-h-h-h:
> Are you aware that you can add multiple items in one go by searching, selecting items, entering a new search query, selecting more items, and so on... The app remembers all selected items from previous searches until you tap the ✔ button.

I didn't know that. That's very useful. So, the gather feature I described [here (end)](https://github.com/davidhealey/waistline/issues/472#issue-1118750223) already exists.

---

To again scroll to the very bottom if that was previously the case, you could establish the following function:

```js
function rememberScrollPosition() {
  const el = $(".page-content");
  const scrollTop = el.scrollTop();

  // Is at very bottom?
  if (scrollTop + el.innerHeight() + 1 >= el.prop("scrollHeight")) { // Can be off by 1.
    app.Diary.lastScrollPosition = Number.MAX_SAFE_INTEGER; // Will scroll to bottom.
  } else {
    app.Diary.lastScrollPosition = scrollTop;
  }
}
```

I used `Number.MAX_SAFE_INTEGER`, since at least outermost `scrollTop` on desktop in Firefox seems to always be an integer and this will most likely keep optimizations in place. It's also described as an integer [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop).

In Firefox, `document.documentElement.scrollTop = Number.MAX_SAFE_INTEGER;` (`0x001f_ffff_ffff_ffff`) on https://www.test.de/ doesn't work. It scrolls to the very top instead. It worked for me in Brave and Opera, which, I guess, use the same engine as Waistline. Strangely, in Firefox, the maximum working value for scrolling to the very bottom is `0x0002_9a43_7fff_ffff` (~8.133 % of `MAX_SAFE_INTEGER`).

h-h-h-h:
> the `+` button I need again right away [@h-h-h-h]

It's also useful hours later for the next meal to have `+` ready on the screen after switching to the app.

PR #458, opened 2022-01-22
Allow linked changes in food editor when creating new food item

### LINKED ISSUE #453: Values not linked when creating new food
When manually creating a new food and having entered all 100-g values, you can't directly gross up to the net package weight. Also, the "link" toggle button at the top is missing. You have to save, edit again and then enter the package weight. This isn't the case when scanning a food (then, you need to enable linking, though, which should be enabled by default).

EmilJunker:
I found a way to make the link button also work when creating new food items. Closes #453.


davidhealey:
Well done!

PR #459, opened 2022-01-22
Allow up to three decimal places for nutrition fields in food editor

### LINKED ISSUE #455: Nutrition value with 3 decimal places invalid
When, for the line "Salt (g)", I try to enter the value "0.046", as it's written on the package, it says "Invalid value."

I think values with many decimal places should always be accepted and saved accurately, even if the presentation thereafter is rounded to 2 decimal places.

EmilJunker:
This was an easy fix :)
Closes #455.


davidhealey:
You're on fire this weekend!

PR #460, opened 2022-01-22
Apply proper number formatting with localized decimal separator

### LINKED ISSUE #456: Proper number formatting
My app instance uses the German locale. That means decimal separator `,` instead of `.`. And also, while leaving out an initial `0` and starting with the decimal separator has a degree of popularity in English, it's totally uncommon in German and looks very strange. My software number keyboard, though, only has `.`, not `,`, and I find it practical to not bother entering the `0` when the machine understands me anyway, so I'll often enter `.5`, e.g.

In the food editor, nutrition values are correctly displayed with `,` and leading `0`.

On the bottom of the diary, fat, carbs and protein are incorrectly displayed with `.`. Same for the statistic section of the app.

In the diary overview, portion counts incorrectly use `.` (like `×.5`). And the display of a leading `0` incorrectly follows whether you entered it when editing the entry. When editing, "Number of Servings" uses `,`, but incorrectly no leading `0` if you didn't enter it in the past.

Separating numbers and units is inconsistent throughout the app. In the diary and the food list, sometimes zero characters (like for `g` and `ml`) and sometimes a space is used (like for `Stück` [= piece(s)]). The right-aligned `kcal` always uses a space. In my view, consistently using a narrow no-break space (U+202F) or a thin space (U+2009), whose merits are the existence of an HTML entity (`&thinsp;`), should be tried.

I also find the uneven spacing surrounding `×` in "400g ×0.5" to be odd. I'd use spaces on both sides.

EmilJunker:
This closes #456 and fixes the formatting issues mentioned there.

### DIFF
--- a/www/activities/foods-meals-recipes/js/foods-meals-recipes.js
+++ b/www/activities/foods-meals-recipes/js/foods-meals-recipes.js
@@ -516,17 +516,21 @@ app.FoodsMealsRecipes = {
         let text = "";
 
         if (item.name != "Quick Add" && item.portion !== undefined && !isNaN(item.portion)) {
-          text = parseFloat(item.portion);
+          text = app.Utils.tidyNumber(parseFloat(item.portion));
 
           if (item.unit !== undefined) {
             if (app.standardUnits.includes(item.unit))
-              text += item.unit;
+              text += "&thinsp;" + item.unit;
             else
               text += " " + app.Utils.escapeHtml(item.unit);
           }
 
-          if (item.quantity !== undefined && item.quantity != 1)
-            text += " &times;" + app.Utils.escapeHtml(item.quantity);
+          if (item.quantity !== undefined && item.quantity != 1) {
+            if ($("html").get(0).getAttribute("dir") === "rtl")
+              text += " &#x202E;&times;&#x202C; " + app.Utils.tidyNumber(parseFloat(item.quantity));
+            else
+              text += " &times; " + app.Utils.tidyNumber(parseFloat(item.quantity));
+          }
         }
 
         if (timestamp == true && item.dateTime !== undefined) {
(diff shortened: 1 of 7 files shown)


www/activities/foods-meals-recipes/js/foods-meals-recipes.js
```diff
@@ -516,17 +516,21 @@ app.FoodsMealsRecipes = {
...
-          text = parseFloat(item.portion);
+          text = app.Utils.tidyNumber(parseFloat(item.portion));
 
           if (item.unit !== undefined) {
             if (app.standardUnits.includes(item.unit))
-              text += item.unit;
+              text += "&thinsp;" + item.unit;
             else
               text += " " + app.Utils.escapeHtml(item.unit);
           }
```
h-h-h-h, line 526: @EmilJunker: Why not just add an optional "unit" parameter to `app.Utils.tidyNumber()` and do this formatting there? It otherwise seems to be a bit inconsistent again.

EmilJunker: Well, it's not that simple. In some languages, e.g. Hebrew, there are special localized versions of certain units. And everything needs to work with a right-to-left layout too. The case distinction for `standardUnits` is really just a thing in the food list to make sure an item with unit "slices" doesn't show up as "2slices" (or "2 slices" with a thin space which also look weird). Everywhere else, it always uses normal spaces. But I like your idea. I think we can indeed make this case distinction part of the tidyNumber function to apply it everywhere.

h-h-h-h: You could even replace and initial hyphen with the proper HTML `&minus;`, if it's for labels only (not textboxes meant to be edited).

PR #468, opened 2022-01-29
Improve diary categories

### LINKED ISSUE #451: Support for not using diary categories
I'm talking about the categories defined under "Settings > Diary > Categories". I have only defined one category for a long time (and I'm hesitant to try what happens if I delete this one). I gave it a generic name in Waisline v2, because it probably didn't work without a category back then. I just don't need these category concepts like "Lunch", because things are not that defined for me.

But, starting with the settings page (a switch there would be good), the app should support not working with categories at all. This would also imply a different rendering of the diary -- one without sections or so. And in the diary entry editor, it saves vertical space and makes everything less cluttered to hide the "Category" line.

BTW: If I'm going to the food list via main menu and add a food to the diary, I'm getting asked "Which meal is this?", where I *have* to tap on my single category name (not even the back button works). That doesn't make sense.

EmilJunker:
This closes #451.

If you only have a single diary category configured, it now hides the category dropdown in the food editor, and it no longer shows the "What meal is this?" prompt.

Also, the diary now renders correctly even if you clear all the categories or leave a gap (like `["Breakfast", "", "Dinner"]`).

While testing this on my phone, I also noticed that the implementation of the scroll position restore feature (#449) in the diary was a bit unreliable. So I made some changes to fix that. It still doesn't always work when animations are enabled, though. The page transition somehow interferes with it. Anyway, I think it's better than nothing.

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -246,22 +246,23 @@ app.Diary = {
       values.id = x + "-value";
 
       let span = document.createElement("span");
-      t = document.createTextNode("0");
+      t = document.createTextNode("");
+      let value = 0;
 
       if (nutrition !== undefined && nutrition[x] !== undefined) {
         if (x !== "calories" && x !== "kilojoules")
-          t.nodeValue = app.Utils.tidyNumber(Math.round(nutrition[x] * 100) / 100);
+          value = (Math.round(nutrition[x] * 100) / 100);
         else
-          t.nodeValue = app.Utils.tidyNumber(Math.round(nutrition[x]));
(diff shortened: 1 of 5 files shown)


davidhealey:
Thanks. I'll merge this. Regarding the scroll position. I think this needs a fundamental change to be implemented properly but for now your fix is exactly what we'll go with.

In Waistline v2 when an item was deleted from the diary it was just removed from the DOM. In v3 I did it differently, I remove it from the database and then reload the diary page. I think going back to the v2 method will provide a better experience, but it's not a critical issue.

PR #476, opened 2022-02-02
Fix OFF upload

### LINKED ISSUE #475: Bug: uploading to OFF zeros out calories
Android version: 10
Waistline version: 3.2.2

What I do:
* Open food list
* Scan a barcode for a rare item
* Receive prompt to add to OFF and choose yes
* Fill out fields including both pictures
* Submit
* Scan barcode again

What I expect:
* See the calorie count that I entered

What happens instead:
* Calories are listed as 0

This has happened with every product I've ever uploaded to OFF using Waistline. I'm getting a bit worried that I'm polluting OFF with bad data!

As always thank you for your work on Waistline David.

EmilJunker:
This should hopefully fix #475.

### DIFF
--- a/www/activities/foodlist/js/open-food-facts.js
+++ b/www/activities/foodlist/js/open-food-facts.js
@@ -218,6 +218,8 @@ app.OpenFoodFacts = {
   getUploadString: function(data) {
     const nutriments = app.nutriments; // Array of OFF nutriment names
     const units = app.nutrimentUnits;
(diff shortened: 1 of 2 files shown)


davidhealey:
That looks good. Have you tested it? If this fixes the problem I'll put out a new release.

EmilJunker:
I couldn't really test it with a real product yet (I only checked if the strings returned by the `getUploadString()` function look alright) but I'm positive this will fix the problem.

PR #487, opened 2022-02-12
Show last applicable values when opening diary log

EmilJunker:
This PR changes the behaviour of the log button in the diary. In the current version, when you press the log button, the dialog gets prefilled with the last values you entered (weight, neck, hips etc.). With the changes in this PR, the dialog now shows the logged values of the selected day instead. If there are no values for the selected day yet, then it checks the previous 14 diary entries for logged values and displays the first one that it finds.


davidhealey:
Very nice!

Issue #501, opened 2022-03-02
Auto-backup file doesn't contain backup data, but nonsense!


h-h-h-h:
Waistline version: 3.3.0

I've enabled the auto-backup feature for a while now. And I just saw that `/storage/emulated/0/Android/media/com.waist.line/waistline_auto_backup.json` contains nothing but `[object Object]` (used `.toString()` instead of `JSON.stringify()`?). The file has a timestamp with the same minute as Waistline was freshly opened.

When exporting manually via button press, `waistline_export.json` from the same directory is 1 megabyte big, so will probably contain an actual backup and valid JSON.

h-h-h-h:
Maybe, as a general extra safety measure, a backup file shouldn't be overwritten, if the length of the new data is *wildly* smaller than the old file's size? Because, if a feature like auto-backup (when working) comes together with a bug like https://github.com/davidhealey/waistline/issues/500, not having overwritten the backup file with quite obviously decimated data, could've saved you. A percentage difference would probably not be the right thing, but a fixed amount of difference. Even if this means ineffectivity for smaller backup file sizes, the bigger files worthier of protection would benefit.

Schroedingberg:
totally down with @h-h-h-h 's proposition. Really happy to see it already merged <3

PR #505, opened 2022-03-02
Include current weekday in auto-backup filename

### LINKED ISSUE #350: Autobackup should have multiple files
Auto backup should alternate between at least two files instead of just one.

EmilJunker:
This closes #350 and should address the concerns raised in https://github.com/davidhealey/waistline/issues/501#issuecomment-1057120468.

The auto-backup now uses filenames like `waistline_backup_wednesday.json`, `waistline_backup_thursday.json` and so on.

### DIFF
--- a/www/index.js
+++ b/www/index.js
@@ -451,7 +451,8 @@ document.addEventListener('deviceready', async function() {
       data.settings = settings;
       let json = JSON.stringify(data);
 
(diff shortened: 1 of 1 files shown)


davidhealey:
Yes I like this solution, and it should naturally limit the number of auto backups to 7 so we don't waste storage.

Issue #410, opened 2021-11-21
Uninstalling app deletes backups


davidhealey:
We need to put the backups in a different folder, android still clears them from the media folder when uninstall the app.

h-h-h-h:
It would be great if a de-facto standard could be established together with other apps to use one common folder for backups and app data not placed in the default system directories. The "root" folder (`/storage/emulated/0`) can get quite crowded. I have these folders from other apps (while there are already enough other ones, I'll never use):

- `gmmp` with app data from [GoneMAD Music Player](https://play.google.com/store/apps/details?id=gonemad.gmmp)
- `Librera` with app data from [Librera](https://play.google.com/store/apps/details?id=com.foobnix.pro.pdf.reader)
- `My Opening Hours` with backup from [My Opening Hours](https://play.google.com/store/apps/details?id=com.meneltharion.myopeninghours.app)
- `Smart AudioBook Player Backup` with backup from [Smart AudioBook Player](https://play.google.com/store/apps/details?id=ak.alizandro.smartaudiobookplayer)

The name would need to be well-thought-out, because it would be permanent.

EmilJunker:
I tried to move the backups into the Documents folder (`/storage/emulated/0/Documents`). This seemed to work at first, and the backup files weren't deleted when I uninstalled the app! However, after I reinstalled it, the app could no longer read (or overwrite) the backup files from the previous installation. So I don't think this is a viable option.

On the plus side, I was able to implement a file picker using https://github.com/cyph/cordova-plugin-chooser that allows me to select and import a backup file from an arbitrary location (and apparently doesn't require any permissions).

davidhealey:
> On the plus side, I was able to implement a file picker using https://github.com/cyph/cordova-plugin-chooser that allows me to select and import a backup file from an arbitrary location (and apparently doesn't require any permissions).

Can this also be used to select a location to store backup files?

EmilJunker:
> Can this also be used to select a location to store backup files?

Unfortunately not. It only lets me select existing files (and not directories).

davidhealey:
Ah that's a shame, but still useful to be able to select the file to load.

PR #515, opened 2022-03-11
Change backup location to Documents directory

EmilJunker:
This is a proposal to solve #410.

Here's what I did:

I changed the backup location to a directory within the Documents folder: `/storage/emulated/0/Documents/Waistline`. Specifically, the first time you use the backup/export feature, Waistline creates a unique backup directory within this folder, e.g. `/storage/emulated/0/Documents/Waistline/1647011571`. All backups and exports are put in this directory (the directory name is just the unix timestamp from the first time you used the backup/export feature, and it is saved in localstorage).

If you ever uninstall Waistline, the backup directory and files are not deleted! 🙌 

If you then reinstall Waistline, the new installation will be unable to overwrite the files created by the previous installation. But since the new installation also creates a new unique backup directory, e.g. `/storage/emulated/0/Documents/Waistline/1647011830`, this is not an issue.

What do you think of this solution?

### DIFF
--- a/www/assets/css/global-styles.css
+++ b/www/assets/css/global-styles.css
@@ -49,6 +49,8 @@ html[dir="rtl"] img.food-thumbnail {margin: 0.2em 0 0.2em 1em;}
 .list .item-title-row .item-after {align-self: flex-start;}
 .list .item-subtitle {font-style: italic;}
 
+.notification-text {overflow-wrap: break-word;}
+
 /* Custom color themes */
 .color-theme-pink {
   --f7-theme-color: #f6b0ba;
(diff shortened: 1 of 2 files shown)


davidhealey:
This looks like an excellent solution, does it work also with the automatic backup if you never make a manual backup?

davidhealey:
I'm unable to select a backup file to restore. It takes me to the file browser but tapping a backup file does nothing.

EmilJunker:
> If I remove the "application/json" filter it works.

Alright, we can just remove then. It's not really important.

Issue #495, opened 2022-02-25
App crashes after taking a picture for OFF upload


rossmeissl:
App version: 3.3.0
Android version: 11

Ever since upgrading to Android 11, Waistline has been exhibiting a hard crash when taking pictures for OFF upload. I used the Android "send feedback" button . . . not sure if those crash reports make it to you or not.

[Screencap of crash](https://photos.app.goo.gl/bN7K3UM2EFbbapHk9)

## What I do

* From the Food screen, tap the barcode icon
* Scan the barcode on a food item that isn't in OFF
* Choose "Yes"
* Tap camera icon
* Take picture of food
* Tap checkmark button to confirm

## What I expect to happen
* Return to data entry screen for OFF upload

## What happens instead
* App crashes

davidhealey:
Nope the feedback didn't reach me.

Does - Settings >> Integration >> Edit images before upload - make any difference?

rossmeissl:
I tried that. An error pops up saying "Editing is not supported for this image." and dumps me back to the "Add New Item" OFF upload screen (with no image).

rossmeissl:
Anything I can do to help here?

EmilJunker:
> Anything I can do to help here? [@rossmeissl]

Since you are currently the only one experiencing this issue, I think you'll have to debug this yourself. The easiest way to do this is by installing a debug APK and then using the browser to inspect the app while it's running on your phone.

If you want to do this, make sure to back up your data first (in Settings > Import/Export > Backup Database) and then move the backup json file to a safe location (the default location will be cleared when you uninstall the app)!

After that, you would need to uninstall Waistline from your phone and install the APK linked below:
[app-debug.zip](https://github.com/davidhealey/waistline/files/8215681/app-debug.zip)

You can then connect your phone to your PC via USB, open any Chromium based browser and go to `chrome://inspect` to see your device listed there and inspect it.

EmilJunker:
@rossmeissl Is this issue still present in the latest release version `3.3.3`?

sheckandar:
I have the same issue on Android 11.

Pixel 3a
App version 3.6.0

The issue happens exactly as OP described.

Using the default Google camera.

nabeelr:
Hi there! Just started having a similar issue, seemingly after the latest update, version 3.11.1.

Though it doesn't seem like a crash, but rather going back to the main screen after showing the Waistline loading brown background seen at first launch again. Usually crashes take you back to the home screen in my experience...

[...]

PR #518, opened 2022-03-13
Fix bug where a meal or recipe cannot be edited

EmilJunker:
This PR fixes a bug where a meal or recipe sometimes cannot be edited after adding an ingredient to it. This was apparently caused by the changes I made in #511.

I also added a changelog file for the next F-Droid release (assuming that will be version `3.3.2`).


davidhealey:
Think it's time for a new release yet?

EmilJunker:
I'm not saying we need to push a new release right now. It would be good if we could fix #495 fist, although I don't really know how.

On the other hand, maybe it would be good to soon release a version where the auto backup is actually working 😄

PR #521, opened 2022-03-18
Fix truncated lines in diary and ingredients lists

EmilJunker:
I see you have released a new version yesterday, but it looks like you didn't update the version code in config.xml, so the release will likely not be picked up by F-Droid. On the plus side, this gives me the opportunity to address one more small bug I noticed where the truncated line fix (#508) didn't work in the diary and in meal/recipe ingredients lists.


davidhealey:
Oh crap, I did update the code, so in the binary it's correct, but I forgot to push it :) I'll do a new one now.

PR #530, opened 2022-03-26
Add ability to set goals with effective from date

### LINKED ISSUE #519: Slight changes in average goal lead to jumps
I have the following settings:

- Same goal every day
- Automatic goal adjustment to meet average goal
- Calculate average goal based on last 10 days

I just reduced my calorie goal by 50 kcal. But the goal in the diary was reduced by several hundred kcal. I wanted the slightly reduced daily goal to apply from today on, not since 10 days ago or whatever the app did, making me pay an added up price. For bigger changes, the goal display must be rendered unusable.

It would be best if the app knew and considered the dates of goal changes and what applied before, at least for the duration of the average calculation.

EmilJunker:
This PR adds the ability to set new nutrition goals without overwriting your previous goals (closes #519).

To do this, I completely reworked the way nutrition goals are stored by the app. Instead of

```json
"goals": {
  "calories": ["2000", "", "", "", "", "", ""],
  "calories-shared-goal": true,
  "calories-show-in-diary": true,
  "calories-show-in-stats": true,
  ...
},
```

you can now have a whole list of goals per nutrient:

```json
"goals": {
  "calories": {
    "show-in-diary": true,
    "show-in-stats": true,
    "goal-list": [{
        "shared-goal": true,
        "goal": ["2000", "", "", "", "", "", ""]
      },
      {
        "effectiveFrom": "2022-03-26T00:00:00.000Z",
        "shared-goal": true,
        "auto-adjust": true,
        "goal": ["2200", "", "", "", "", "", ""]
    }]
  },
  ...
},
```

The goals in the list may have an `effectiveFrom` date that is used to determine which goal should be used on a given day.

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -237,7 +237,7 @@ app.Diary = {
     let goals = await app.Goals.getGoals(nutrimentsToShow, date);
 
     nutrimentsToShow.forEach((x) => {
-      let goal = goals[x];
+      let nutrimentGoal = goals[x];
 
       // Show n nutriments at a time
       if (count % columnsToShow == 0) {
@@ -265,7 +265,7 @@ app.Diary = {
       title.appendChild(t);
       rows[0].appendChild(title);
 
-      // Values and goal text
+      // Value
       let values = document.createElement("div");
       values.className = "col keep-ltr";
       values.id = x + "-value";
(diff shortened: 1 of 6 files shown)


davidhealey:
My only suggestion would be to change effectiveFrom to effective-from to maintain consistency with the other object keys. Although I prefer camel case so I have no idea why I went the `-` route to begin with (perhaps I was copying something else) :)

EmilJunker:
> change effectiveFrom to effective-from

Oh, I didn't even notice. Sure, I can do that.

EmilJunker:
> change effectiveFrom to effective-from

Done

PR #531, opened 2022-03-26
Improve integration of OFF and USDA food databases

### LINKED ISSUE #524: Use item units (1 package, 2 slices, 1 banana) instead of 100 g by default for USDA items
When scanning a barcode, or searching for an item, almost all results are displayed with a serving size of 100 g.

This makes for a mild annoyance when scanning a barcode, because I'll need to press the lock button for scaling, enter the calories on the package to scale all the nutrient numbers, then unlock it to change the unit to something sensible such as 1 package or 1 cup. For instance, I scanned a carton of milk, and it gave me 66.667 kcal for a "serving size" of 100 g. The correct numbers are 160 kcal for a serving size of 1 cup (or 240 mL).

Part of this might be a problem with Open Food Facts). It seems to be an issue primarily with US food items added by the user "kiliweb" which is always then edited by "Yuka" (which is an app). These users/app only add 100g and never add the serving size. Not sure if openfoodfacts is aware of this.

However, I know the USDA database has the item units, but waistline does not use them. For instance, searching "Bananas, overripe, raw" will give only the value for 100 g, even though the [USDA page](https://fdc.nal.usda.gov/fdc-app.html#/food-details/1105073/nutrients) for this item has a dropdown that let's you use "1 banana" as a unit. For generic food items, item units are more essential since, without actually weighing the item, there is no label to use to scale the numbers to the correct value.

### LINKED ISSUE #525: OFF serving sizes with fractions (a / b) are not read correctly
[This item](https://world.openfoodfacts.org/product/0030000063545/cereal-quaker) has a serving size of "3/4" and a serving unit of "CupCerealAlone". However, in the Waistline app, the serving size is given as "3" (the /4 is not read).

EmilJunker:
This PR improves various aspects of the OFF and USDA integration.

In particular, Waistline now respects the serving sizes and item units of many USDA foods as I outlined in https://github.com/davidhealey/waistline/issues/524#issuecomment-1079668428 (closes #524).

Also, serving sizes with fractions are now recognized and parsed properly (closes #525).

When re-downloading the information for a food item, the user can now choose between downloading it per serving or per 100g.

In the food editor, there is now a direct link to the USDA product page of an item if it was imported from there (i.e. if it has a barcode that starts with "fdcId_").


EmilJunker:
> I'll let it sit for a few days and see if anyone reports any issues

I'll install it on my phone and play around with it over the next couple of days. If anyone else wants to test the new features, here is an APK you can use: [app-debug.zip](https://github.com/davidhealey/waistline/files/8356033/app-debug.zip). Just make sure to backup your data first because you'll need to uninstall the current version to be able to install this one.

PR #539, opened 2022-04-03
Prevent adding meals and recipes when coming from the diary

### LINKED ISSUE #538: Can't add food to meals from diary menu
I can't add food to meals from the diary menu. Not sure if this was always the case, but seems like a flaw. You can add food to meals if you go directly to the foods menu (not through the diary menu)
Have included a video below.

Edit: am using version 3.3.3

https://user-images.githubusercontent.com/9134591/161380270-e4f0babf-e862-4186-937d-83b7938f98cf.mp4

EmilJunker:
This is a follow-up to #511 to clear up some of the confusion as discussed in #538.
Edit: closes #538.

### DIFF
--- a/www/activities/foodlist/js/foodlist.js
+++ b/www/activities/foodlist/js/foodlist.js
@@ -33,8 +33,7 @@ app.Foodlist = {
     this.getComponents();
     this.createSearchBar();
     this.bindUIActions();
-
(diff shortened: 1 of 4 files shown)


davidhealey:
Thanks! I'm going to put out a new release later today.

PR #578, opened 2022-04-30
Minor usability improvements

EmilJunker:
This PR contains some minor usability improvements.

Apart from this, I haven't noticed any major bugs or problems with the new features. Admittedly, the image compression is giving mixed results. Sometimes the resulting images are a bit pixilated, especially when selecting existing pictures from the gallery. I assume that's because they are already compressed and can't be made much smaller without reducing the quality.

Here is an example: Before the compression...

photo_before

After the compression...

photo_after

On the other hand, most of the time the compressed image looks perfectly fine and is barely distinguishable from the original:
Before the compression...

apples_before

After the compression...

apples_after

Anyway, I haven't found a better solution for the compression yet, so I'd say we go ahead with the release and see how well it works in practise. This way, we can also let more people test it on different phones/hardware.

### DIFF
--- a/www/activities/foods-meals-recipes/js/food-editor.js
+++ b/www/activities/foods-meals-recipes/js/food-editor.js
@@ -549,9 +549,6 @@ app.FoodEditor = {
 
       fetch(uri).then((res) => res.blob()).then(async (blob) => {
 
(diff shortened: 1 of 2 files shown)


davidhealey:
Excellent work! I'll push a release later today.

PR #580, opened 2022-05-07
Fix crop photos option in food editor

### LINKED ISSUE #579: Remove edit image option
The edit image option doesn't work at all with my phone since we updated to the latest cordova camera plugin. This seems to be a known issue with no proper solution.

I suggest we remove the edit image setting and disable that option in the takePicture function. We should then use cordova-plugin-crop to provide the cropping capability. https://github.com/jeduan/cordova-plugin-crop

I did a little test earlier today and this plugin seems to work nicely with file uri but not base64. So I suggest we go back to file uri and convert the cropped image to base64.

I don't have time to tackle this at the moment unfortunately but will try to get round to it in the next few weeks. @EmilJunker If you have time and want to give it a go please feel free.

EmilJunker:
This PR implements the image cropping functionality and closes #579.

I'm using the crop plugin from https://github.com/rastafan/cordova-plugin-crop instead of the original because this one allows to freely change the aspect ratio.

### DIFF
--- a/metadata/en-US/changelogs/30402.txt
+++ b/metadata/en-US/changelogs/30402.txt
@@ -1,0 +1,5 @@
+- fixed the option to crop photos in the food editor
+- various bugfixes
+
+You can find the full changelog at 
+https://github.com/davidhealey/waistline/releases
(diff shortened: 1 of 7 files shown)


davidhealey:
Nice work. I like the fileToBlob function!

PR #593, opened 2022-06-21
Add ability to clone items and to show and restore archived items

### LINKED ISSUE #552: Question about recipes and tracking
I am mostly using recipes to track my daily calories since they can be scaled (say, 1/2 of the recipe because I did share the meal with soneome).

So far so good.

But whenever I adapt a recipe all diary entries will be adapting. Resulting in completely wrong statistics. Which makes them useless for me.

Am I doing sonething wrong?
**How can you scale meals and have correct statistics?**

Thanks in advance!

version: 3.4.0

EmilJunker:
This PR implements the changes proposed in https://github.com/davidhealey/waistline/issues/552#issuecomment-1102168935 and closes #552.

When you long press a food, meal or recipe, you are now presented with multiple options, including the ability to clone the item:

image1

You can also browse archived foods/recipes via the filter menu to edit or restore them:

image2

Note that I had to drop the uniqueness constraint from the barcode index of the foodlist table. This was necessary to allow multiple entries of the same product to coexist (in case the product nutrition information changes).


davidhealey:
Oh excellent!

PR #600, opened 2022-07-09
Add option to follow the system wide dark theme (Android 10+)

### LINKED ISSUE #463: Automatic dark theme for Android 10+
Most modern applications support an option to follow the system wide dark theme. It would be greatly appreciated if you could make it.

EmilJunker:
This closes #463. I even managed to implement the feature in a way that doesn't require an app restart.

To make the theme detection work, I had to add this plugin:
https://github.com/timbru31/cordova-plugin-android-dark-mode-support

The plugin is needed because the webview used by cordova-android apparently doesn't support the (prefers-color-scheme) media query by default. I don't really understand why. But anyway, it works with the plugin.

By the way @davidhealey are you serious about this "switching away from GitHub" stuff? I just saw that you edited the readme a few days ago. I haven't been following this whole controversy too actively, and I honestly don't understand what all the fuzz is about. After all, GitHub is offering their services completely free of charge, including unlimited repo hosting, CI/CD, and so much more. I feel like that's a pretty sweet deal. Maybe you can enlighten me :)

### DIFF
--- a/www/assets/locales/locale-en.json
+++ b/www/assets/locales/locale-en.json
@@ -219,7 +219,10 @@
     "needs-restart": "Restart app to apply changes.",
     "appearance": {
       "title": "Appearance",
-      "dark": "Dark Mode",
+      "app-mode": "App Mode",
+      "system": "Use System Theme",
+      "light": "Light",
+      "dark": "Dark",
       "colour-theme": "Colour Theme",
       "animations": "Disable UI Animations",
       "start-page": "Start Page",
--- a/config.xml
+++ b/config.xml
@@ -20,6 +20,8 @@
         <preference name="AndroidPersistentFileLocation" value="Compatibility" />
         <preference name="AndroidExtraFilesystems" value="files,sdcard,cache" />
         <preference name="AndroidInsecureFileModeEnabled" value="true" />
+        <preference name="AndroidXEnabled" value="true" />
+        <preference name="GradlePluginKotlinEnabled" value="true" />
         <config-file after="uses-permission" parent="/manifest" target="AndroidManifest.xml">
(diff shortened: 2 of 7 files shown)


davidhealey:
This looks good, well done!

Yes I've been intending for sometime to host my own git instance for my repos (probably with a clone to another service as well).

My desire to move away from github was greatly amplified by Microsoft taking over and again when co-pilot was announced. The idea of my code potentially being used without permission in a proprietary project really annoys me. The fact that github is free of charge doesn't make a difference to me. I'd happily pay for a service that gave the same features without the abuse of the user's data.

www/assets/locales/locale-en.json
```diff
@@ -219,7 +219,10 @@
     "needs-restart": "Restart app to apply changes.",
     "appearance": {
       "title": "Appearance",
-      "dark": "Dark Mode",
+      "app-mode": "App Mode",
```
janibonnevier, line 222: I was just about to translate this, but found this title a bit peculiar. It's vague enough that I had to come here to double check what it was. Wouldn't "Theme" be more to the point?

davidhealey: The theme is the colours the UI will use, but each theme has two variations/modes, dark and light.

EmilJunker: Yeah I admit App Mode is not an ideal name. But I thought the three available options `Use System Theme`, `Light`, and `Dark` should make it clear what it is. Do you have an idea for a better name?

davidhealey: Maybe theme mode would be better.

EmilJunker: Yes, Theme Mode sounds good. Go ahead and change it if you want. I don't want to make a PR just for that ;)

PR #623, opened 2022-09-10
feature to send data to external service

Kroseida:
I am not familar with the code.
And this is also my first cordova project.

So please give me some feedback or ideas to do it better. :)

What does this feature?
This feature will enable users to share data with their local services. (e.g insert the data into a database)

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -179,7 +179,6 @@ app.Diary = {
 
     // Populate groups and get overal nutrition
     if (entry) {
-      await this.populateGroups(entry);
       let totalNutrition = await app.FoodsMealsRecipes.getTotalNutrition(entry.items, "ignore");
 
       let entryDetails = await Promise.all(entry.items.map(async (data) => {
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -171,8 +171,11 @@ app.Diary = {
   },
 
   sendStatistics: async function () {
-    let address = app.Settings.get("statistics", "send-data-address");
-    if (app.Settings.get("statistics", "send-data") == true && !!address) {
+    let address = app.Settings.get("developer", "data-sharing-address");
+    let wifiOnly = app.Settings.get("developer", "data-sharing-wifi-only");
+
+    if (app.Settings.get("developer", "data-sharing-active") == true && !!address
+        && (wifiOnly && navigator.connection.type == "wifi") || !wifiOnly) {
       let entry = await this.getEntryFromDB(); // Get diary entry from DB
       if (entry) {
         let totalNutrition = await app.FoodsMealsRecipes.getTotalNutrition(entry.items, "ignore");
@@ -185,7 +188,7 @@ app.Diary = {
         await app.Utils.timeoutFetch(address, {
           headers: {
             "User-Agent": "Waistline - Android - Version " + app.version + " - https://github.com/davidhealey/waistline",
-            "Authorization": app.Settings.get("statistics", "send-data-authorization")
+            "Authorization": app.Settings.get("developer", "data-sharing-authorization")
           },
           method: 'POST',
           body: JSON.stringify({"nutrition": totalNutrition, "entryDetails": entryDetails, "entry": entry})
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -174,25 +174,26 @@ app.Diary = {
     let address = app.Settings.get("developer", "data-sharing-address");
     let wifiOnly = app.Settings.get("developer", "data-sharing-wifi-only");
 
-    if (app.Settings.get("developer", "data-sharing-active") == true && !!address
-        && (wifiOnly && navigator.connection.type == "wifi") || !wifiOnly) {
-      let entry = await this.getEntryFromDB(); // Get diary entry from DB
-      if (entry) {
-        let totalNutrition = await app.FoodsMealsRecipes.getTotalNutrition(entry.items, "ignore");
+    if (app.Settings.get("developer", "data-sharing-active") == true && !!address) {
+      if ((wifiOnly && navigator.connection.type == "wifi") || !wifiOnly) {
+        let entry = await this.getEntryFromDB(); // Get diary entry from DB
+        if (entry) {
+          let totalNutrition = await app.FoodsMealsRecipes.getTotalNutrition(entry.items, "ignore");
 
-        let entryDetails = await Promise.all(entry.items.map(async (data) => {
-          return await app.FoodsMealsRecipes.getItem(data.id, data.type, data.portion, data.quantity);
-        }));
+          let entryDetails = await Promise.all(entry.items.map(async (data) => {
+            return await app.FoodsMealsRecipes.getItem(data.id, data.type, data.portion, data.quantity);
+          }));
 
-        // Send nutrition and diary to target service
-        await app.Utils.timeoutFetch(address, {
-          headers: {
-            "User-Agent": "Waistline - Android - Version " + app.version + " - https://github.com/davidhealey/waistline",
-            "Authorization": app.Settings.get("developer", "data-sharing-authorization")
-          },
-          method: 'POST',
-          body: JSON.stringify({"nutrition": totalNutrition, "entryDetails": entryDetails, "entry": entry})
-        });
+          // Send nutrition and diary to target service
+          await app.Utils.timeoutFetch(address, {
+            headers: {
+              "User-Agent": "Waistline - Android - Version " + app.version + " - https://github.com/davidhealey/waistline",
+              "Authorization": app.Settings.get("developer", "data-sharing-authorization")
+            },
+            method: 'POST',
+            body: JSON.stringify({"nutrition": totalNutrition, "entryDetails": entryDetails, "entry": entry})
+          });
+        }
       }
     }
   },
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -170,27 +170,29 @@ app.Diary = {
     el.innerText = dateString;
   },
 
-  sendStatistics: async function() {
+  sendStatistics: async function () {
+    if (app.Settings.get("statistics", "send-data") != true) {
+      return;
+    }
+
     let entry = await this.getEntryFromDB(); // Get diary entry from DB
-    let totalNutrition;
 
     // Populate groups and get overal nutrition
     if (entry) {
       await this.populateGroups(entry);
-      totalNutrition = await app.FoodsMealsRecipes.getTotalNutrition(entry.items, "ignore");
-    }
-    let entryDetails = await Promise.all(entry.items.map(async (data) => {
-      return await app.FoodsMealsRecipes.getItem(data.id, data.type, data.portion, data.quantity);
-    }));
+      let totalNutrition = await app.FoodsMealsRecipes.getTotalNutrition(entry.items, "ignore");
 
-    if (app.Settings.get("statistics", "send-data") == true && totalNutrition) {
+      let entryDetails = await Promise.all(entry.items.map(async (data) => {
+        return await app.FoodsMealsRecipes.getItem(data.id, data.type, data.portion, data.quantity);
+      }));
+
       // Send nutrition and diary to target service
       await app.Utils.timeoutFetch(app.Settings.get("statistics", "send-data-address"), {
         headers: {
           "User-Agent": "Waistline - Android - Version " + app.version + " - https://github.com/davidhealey/waistline"
         },
         method: 'POST',
-        body: JSON.stringify({ "nutrition": totalNutrition, "entryDetails": entryDetails, "entry": entry })
+        body: JSON.stringify({"nutrition": totalNutrition, "entryDetails": entryDetails, "entry": entry})
       });
     }
   },
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -53,7 +53,6 @@ app.Diary = {
 
     if (document.querySelector(".page-current[data-name='diary']") != null)
       app.Diary.lastScrollPosition = 0; // Reset last scroll position
-
   },
 
   getComponents: function() {
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -53,6 +53,7 @@ app.Diary = {
 
     if (document.querySelector(".page-current[data-name='diary']") != null)
       app.Diary.lastScrollPosition = 0; // Reset last scroll position
+
   },
 
   getComponents: function() {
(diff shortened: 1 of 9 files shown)


www/activities/diary/js/diary.js
```diff
@@ -220,6 +245,9 @@ app.Diary = {
     if (requestedScrollPosition !== undefined) {
       $(".page[data-name='diary'] .page-content").scrollTop(requestedScrollPosition);
     }
+
+    // Send statistics. This is the wrong place. but idk. a better place.
+    this.sendStatistics();
```
davidhealey, line 250: This should probably be an action in the import/export page

www/activities/diary/js/diary.js
```diff
@@ -170,6 +170,31 @@ app.Diary = {
     el.innerText = dateString;
   },
 
+  sendStatistics: async function() {
+    let entry = await this.getEntryFromDB(); // Get diary entry from DB
+    let totalNutrition;
+
+    // Populate groups and get overal nutrition
+    if (entry) {
+      await this.populateGroups(entry);
```
davidhealey, line 179: Why do the groups need to be populated here?

www/activities/settings/views/statistics.html
```diff
@@ -72,6 +72,30 @@
             </div>
           </div>
         </li>
+
+        <li>
+          <div class="item-content">
+            <div class="item-inner">
+              <div class="item-title" data-localize="settings.statistics.send-data">Send data to own service</div>
```
davidhealey, line 79: I think this could be worded better. Perhaps "Send data to external service"

davidhealey (changes requested):

davidhealey:
Hey this looks pretty good, thanks! I've added a few comments. @EmilJunker Could you also take a look at this to see if there is anything important I've missed. You're more familiar with the codebase than I am at the moment.

www/activities/diary/js/diary.js
Kroseida: "sendStatistics" will send the data to the external service.

With moving it into import/export page it will only sent it to the external service on loading the import / export page.

www/activities/settings/views/statistics.html
Kroseida: done

www/activities/diary/js/diary.js
Kroseida: done

www/activities/diary/js/diary.js
EmilJunker: Placing the function here will execute it every time a diary item is added, removed, or edited, but also when you just open the diary page. So this is probably not the behaviour you want.
To fix this, see where the render function is called from and consider if adding your function call there makes sense.

EmilJunker: I would suggest to place the "sendStatistics" function calls in these places:
- https://github.com/Kroseida/waistline/blob/6df9f359273325f6e239a0a6e7df0241d95f7e69/www/activities/diary/js/diary.js#L517
- https://github.com/Kroseida/waistline/blob/6df9f359273325f6e239a0a6e7df0241d95f7e69/www/activities/diary/js/diary.js#L618
- https://github.com/Kroseida/waistline/blob/6df9f359273325f6e239a0a6e7df0241d95f7e69/www/activities/diary/js/diary.js#L641
- https://github.com/Kroseida/waistline/blob/6df9f359273325f6e239a0a6e7df0241d95f7e69/www/activities/diary/js/diary.js#L677
- https://github.com/Kroseida/waistline/blob/6df9f359273325f6e239a0a6e7df0241d95f7e69/www/activities/diary/js/diary.js#L765

www/activities/diary/js/diary.js
```diff
@@ -170,6 +170,33 @@ app.Diary = {
     el.innerText = dateString;
   },
 
+  sendStatistics: async function () {
+    if (app.Settings.get("statistics", "send-data") != true) {
+      return;
```
EmilJunker, line 175: Instead of leaving the function when this setting is `!= true`, I think it would be better and cleaner to only go into the function if it is `== true`. But this is a matter of style, of course.

www/activities/diary/js/diary.js
```diff
@@ -170,6 +170,33 @@ app.Diary = {
...
+    }
+
+    let entry = await this.getEntryFromDB(); // Get diary entry from DB
+
+    // Populate groups and get overal nutrition
```
EmilJunker, line 180: This comment here is not a good description of what the code does. Consider changing or removing it.

www/activities/diary/js/diary.js
```diff
@@ -170,6 +170,33 @@ app.Diary = {
...
+    if (entry) {
+      let totalNutrition = await app.FoodsMealsRecipes.getTotalNutrition(entry.items, "ignore");
+
+      let entryDetails = await Promise.all(entry.items.map(async (data) => {
+        return await app.FoodsMealsRecipes.getItem(data.id, data.type, data.portion, data.quantity);
+      }));
+
+      // Send nutrition and diary to target service
+      await app.Utils.timeoutFetch(app.Settings.get("statistics", "send-data-address"), {
```
EmilJunker, line 189: I haven't tested this, bit I think this call would result in an error if the `send-data-address` is not set by the user. Consider checking whether the address is set before making this call.

www/activities/diary/js/diary.js
```diff
@@ -170,6 +170,33 @@ app.Diary = {
...
+        headers: {
+          "User-Agent": "Waistline - Android - Version " + app.version + " - https://github.com/davidhealey/waistline",
+          "Authorization": app.Settings.get("statistics", "your_authorization_header")
```
EmilJunker, line 192: I believe you need to change `your_authorization_header` to `send-data-authorization` here.

www/activities/settings/views/statistics.html
```diff
@@ -72,6 +72,40 @@
...
+              </div>
+              <div class="item-input-wrap">
+                <input type="text" name="send-data-address" field="statistics" value="https://your.service.org/waistline/">
+              </div>
+            </div>
+          </div>
+          <div class="item-content item-input">
+            <div class="item-inner">
+              <div class="item-title item-label">
+                <span data-localize="settings.statistics.send-data-authorization">Service Authentication</span>
```
EmilJunker, line 101: Consider changing this string to "Service Authentication Header" to make it more clear what it is.

www/activities/settings/views/statistics.html
```diff
@@ -72,6 +72,40 @@
...
+              </div>
+              <div class="item-input-wrap">
+                <input type="text" name="send-data-authorization" field="statistics" value="">
```
EmilJunker, line 104: This should be `placeholder="https://your.service.org/waistline/"` instead of `value="https://your.service.org/waistline/"`.

www/activities/settings/views/statistics.html
```diff
@@ -72,6 +72,40 @@
...
+              </div>
+            </div>
+          </div>
+          <div class="item-content item-input">
+            <div class="item-inner">
+              <div class="item-title item-label">
+                <span data-localize="settings.statistics.send-data-address">Service Address</span>
+              </div>
+              <div class="item-input-wrap">
+                <input type="text" name="send-data-address" field="statistics" value="https://your.service.org/waistline/">
```
EmilJunker, line 94: This should be `placeholder="https://your.service.org/waistline/"` instead of `value="https://your.service.org/waistline/"`.

www/activities/settings/views/statistics.html
```diff
@@ -72,6 +72,40 @@
             </div>
           </div>
         </li>
+
+        <li>
+          <div class="item-content">
+            <div class="item-inner">
+              <div class="item-title" data-localize="settings.statistics.send-data">Send data to external service</div>
```
EmilJunker, line 79: You should add the locale strings to the `locale-en.json` file ([here](https://github.com/Kroseida/waistline/blob/6df9f359273325f6e239a0a6e7df0241d95f7e69/www/assets/locales/locale-en.json#L293)) so they can be translated by other users.

www/activities/diary/js/diary.js
Kroseida: Exactly this was the purpose I want.
My problem is just the point: sending data in a render function is a little bit weird.

But user should be able to send data to their service by just opening the page, so they can just "sync" old data.
The Backend Part / Service will handle everything else and make Duplications to updates.
I can also change as you suggested. But I would recommend to send the data on every action. (Also loading the page.)
So you can also sync the data if you have no internet connection while adding entries.

Pro:
- Better Data Sync (also for old data)

Con:
- More Requests

Offtopic: I am currently working on the service part as an open-source solution. I could push it to my personal github page or if you have an group I could also push it in your group.

EmilJunker: That reminds me, it would probably be a good idea to check whether the device is connected to the internet before calling the "sendStatistics" (or checking it inside the function). And maybe it should even only send the data when connected to Wi-Fi to save mobile bandwidth.

www/activities/settings/views/statistics.html
Kroseida: done

www/activities/settings/views/statistics.html
Kroseida: done

www/activities/settings/views/statistics.html
Kroseida: changed to "Service Authorization Header"

www/activities/diary/js/diary.js
Kroseida: ohh. yes. done.

www/activities/diary/js/diary.js
Kroseida: added check for this part. Tested in browser - app does not crash or smth.

www/activities/diary/js/diary.js
Kroseida: true

www/activities/diary/js/diary.js
Kroseida: jeah. its my style. I still changed it to "== true".

www/activities/diary/js/diary.js
Kroseida: maybe this should be a setting entry too?

EmilJunker: I think for the foreseeable future, you will be the only person using this feature. So decide for yourself whether you need a setting or not :D
But since the app will send data to the service every time you open the diary, only sending when connected to Wi-Fi should be the default, in my opinion. Otherwise this would use up too much mobile data.

Kroseida: definitely a good point! (This feature is for me and some friends. So we have some more ppl. :P)
I will implement this point as a setting entry. And make it to default true.

EmilJunker: > Maybe we should move this entire section from "statistics" to "developer features"?

Yes, that sounds like a good idea to me.

davidhealey:
Sending the data every time the diary is opened seems a bit excessive. What about just each time the app is opened?

www/activities/diary/js/diary.js
Kroseida: done: (example)
image

defaults are:
        "data-sharing-active": false,
        "data-sharing-wifi-only": true,
        "data-sharing-address": "",
        "data-sharing-authorization": ""

@EmilJunker

Kroseida:
not sure. Maybe this could not be enough? 
Some ppl. could run the app in the background, and never restart it.

And we need to know which day to send exactly. Sending all days would make much trouble.
Sending last day could also make problems.

davidhealey:
I didn't check but can this be completely disabled?

Kroseida:
wdym? Data Sharing is by default disabled. And by default no data will be exposed or sent anywhere.
Just by enabling "Share data with external service" and entering an URL you will get impacted by functionality.

www/activities/diary/js/diary.js
EmilJunker: > Wifi could not be tested. I have to learn how to build it for android

I didn't test it either, but the source code looks good to me, so it should work :)

davidhealey:
@EmilJunker You think it's ready for merging?

www/assets/locales/locale-en.json
```diff
@@ -294,10 +294,14 @@
...
       "y-zero": "Begin Y Axis at Zero",
       "average-line": "Show Average Line",
-      "goal-line": "Show Goal Line",
-      "send-data": "Send data to external service",
-      "send-data-address": "Service Address",
-      "send-data-authorization": "Service Authorization Header"
+      "goal-line": "Show Goal Line"
+    },
+    "developer": {
+      "data-sharing": "Send data to external service",
```
EmilJunker, line 300: A `title` string is missing here.

EmilJunker:
I think this pretty much ready to be merged. There are only a few things left to do:
- add the missing `title` string in `locale-en.json`
- remove the comment that says this is the wrong place to call the function (and if you feel like it's necessary, maybe just rename the "render" function to something else)
- add a line to the [changelog file](https://github.com/davidhealey/waistline/blob/master/metadata/en-US/changelogs/30405.txt) that briefly describes this new feature

www/assets/locales/locale-en.json
Kroseida: I will fix this tomorrow. I will also test this version on Android Studio and remove the "WIP" flag.

Kroseida:
Done. Ready to merge @davidhealey @EmilJunker

www/activities/diary/js/diary.js
```diff
@@ -170,6 +170,33 @@ app.Diary = {
     el.innerText = dateString;
   },
 
+  sendStatistics: async function () {
+    let address = app.Settings.get("developer", "data-sharing-address");
+    let wifiOnly = app.Settings.get("developer", "data-sharing-wifi-only");
+
+    if (app.Settings.get("developer", "data-sharing-active") == true && !!address
```
EmilJunker, line 177: This condition expression is incorrect! Since `&&` binds stronger than `||`, the statement evaluates to `true` even when data sharing is disabled and no address is set. It suffices that `wifiOnly` is `false` for the whole statement to become `true`.

To fix this and to make the code more readable, I'd suggest to change it to:

```js
if (app.Settings.get("developer", "data-sharing-active") == true && !!address) {
  if ((wifiOnly && navigator.connection.type == "wifi") || !wifiOnly) {
    ...
  }
}
```

www/activities/settings/views/statistics.html
```diff
@@ -72,6 +72,7 @@
             </div>
           </div>
         </li>
+
```
EmilJunker, line 75: Please remove this empty line. This seems to be a leftover from an earlier change.

www/activities/diary/js/diary.js
Kroseida: will change it. But I guess this worked because of the double check. but not sure.
should not work tbh.

because of this I normaly work like
```
if (app.Settings.get("developer", "data-sharing-active") == false) {
   return;
}
if(!address) {
   return;
}
if (navigator.connection.type == "wifi" || !wifiOnly) {
    return;
}
```
In my opinion its better for understanding.
But I will change it to your suggestions.

Kroseida: done

www/activities/settings/views/statistics.html
Kroseida: done

Kroseida:
Ready to merge @davidhealey @EmilJunker

davidhealey:
Looks good to me but I'd like @EmilJunker opinion before I merge

EmilJunker:
LGTM 👍

EmilJunker:
@Kroseida The data sharing feature has now been released as part of Waistline version 3.5.0. How is your backend service implementation coming along? It would be great if you could make the code for it public so other people can use it too.

hippothomas:
Since the original backend project was abandoned, I made my own service to centralize all my statistics from the Waistline app. If anyone needs this, feel free to use it: [hippothomas/waistline-api](https://github.com/hippothomas/waistline-api)

PR #634, opened 2022-09-22
Added "Blood Glucose" and "Ketones" tracking.

Helspank:
Tracking Blood Glucose (Glycemia) and Ketones is very useful if you're on a keto diet and/or an Intermittent Fasting regimen, or if you suffer from diabetes.


EmilJunker:
Instead of bloating the interface with fields only a small fraction of users care about, I think it would be better to add support for custom body stats, similar to the custom nutrition fields that are already supported. This way, every user could define their own set of body parameters to track.

Helspank:
> every user could define their own set of body parameters to track.
I like this. Also support for custom measurement units.

Issue #637, opened 2022-09-25
Enhancement suggestion: Visualize tracking gaps in Statistic Chart


schwebke:
Currently gaps in the data, for example from vacations, diet breaks or simply missing out, are not visible in the statistic chart.
Having these improves the ability to interpret the data by keeping the time axis linear.
See screenshots below for how this would look like.

If you feel this of general interest I am happy to contribute a PR.

_current chart_
![current chart](https://user-images.githubusercontent.com/6435375/192144692-7916aa07-1c33-4171-8975-02ebb57a2e70.jpg)

_chart with visualized gaps_
![chart with visualized gaps](https://user-images.githubusercontent.com/6435375/192144728-a5b04900-e9ff-4edb-b97e-3e4c173489d8.jpg)

davidhealey:
Looks like a good idea to me

PR #638, opened 2022-09-25
Visualize tracking gaps in Statistic Chart

schwebke:
Implements #637 

#### Changes
- retain diary data without values
- check and fill in missing days between diary entries
- enable _spanGaps_ option for line chart -- line chart renders an interpolated view; bars show gaps prominently


schwebke:
Thanks for taking a look. Same applies for a gap at the end.
I added trimming from both sides.

PR #645, opened 2022-10-08
Hide number of servings field for recipes

### LINKED ISSUE #642: Recipe editor should say # of servings
The recipe creator has a field for serving size but it should have a field for number of servings instead. Or alternatively should have both Serving Size & # of Servings. As it is currently I have been putting # of Servings in Serving Size field then when adding the food to diary adjust Serving Size field back down to 1.

EmilJunker:
This closes #642. The "Number of Servings" field in the diary is now hidden for recipes. This should clear up the confusion about the issue.

Adding a recipe to the diary via the quantity prompt:
Before...
localhost_

After...
localhost_ (1)

Viewing/Editing the diary entry of a recipe:
Before...
localhost_ (2)

After...
localhost_ (3)

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -555,6 +555,8 @@ app.Diary = {
         ["serving-size", "number-of-servings"].forEach((field) => {
           let li = document.createElement("li");
           li.className = "item-content item-input";
+          if (field == "number-of-servings" && item.type == "recipe")
+            li.style.display = "none"; // number of servings field not needed for recipes
           ul.appendChild(li);
 
           let inner = document.createElement("div");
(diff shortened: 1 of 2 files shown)


pablob127:
This actually broke down one of the ways I was using to enter recipes!
For instance, I have a soup recipe where I have measured that the total quantity is 1350 ml, so I put 1350 ml as the serving size in the recipe. Then, when we share the soup equally between 4 people, I would just enter 1350 ml in the serving, and then 0.25 servings in the number of servings  so that the app would do the math.
With the hidden number of servings field, I need to do the math manually to enter the right amount in the serving size.
I would appreciate it if this change could be made optional (or another way to fix the original issue was found).

PR #650, opened 2022-11-11
Add support for custom body stats

EmilJunker:
This closes #634.

Users can now customize the set of body stats they want to track. They can add their own fields to the list, change the order, and also disable fields they don't care about.

Turning off all fields will now hide the log button on the diary page (unlike before where the weight field was always visible and couldn't be turned off).

@davidhealey I think we have enough new features and fixes to push a new realease soon. I've already prepared the changelog files for F-Droid. It's your choice whether to call this version `3.4.5` or make the jump to `3.5.0`.

localhost_1  localhost_2


davidhealey:
Let's go with 3.5.0

Issue #412, opened 2021-11-28
Why does Waistline import ingredients and traces from OFF?


EmilJunker:
For a while now, when importing food items from OFF, Waistline also imports the `ingredients_text` and `traces` fields. I would be interested to know why it does that. As far as I can tell, this information isn't used for anything. But especially the `ingredients_text` can take up a lot of valuable storage space in the local database.

If this is about food intolerances, I think it would make more sense to import just the `allergens` field instead of the whole `ingredients_text`.

Otherwise, maybe we can drop these fields altogether?

davidhealey:
> Otherwise, maybe we can drop these fields altogether?

Yep let's drop them. I think I added support for importing them when I added support for uploading them, but can't really see a use for it. I don't think we need allergens either.

Might also be good to remove the uploading of ingredients and traces too since I think the app should keep it's focus on calorie/nutrition tracking.

teolemon:
@davidhealey Allowing Waistline users to at least take a photo of the ingredients would be hugely beneficial to the project. Wording it like "Help Waistline's datasource by sending a photo of the ingredients" could set expectations right.

davidhealey:
Maybe we could have an additional images section.

PR #665, opened 2023-01-07
Add ability to upload additional photos to Open Food Facts

EmilJunker:
This adds the ability to take additional photos when uploading items to OFF, as suggested in #412.

In addition to the product photo and nutrition photo, you can now upload an ingredients photo and a packaging photo. Of course, this is entirely optional, the upload also works without the photos.

Also, after the upload is complete, you now get asked to immediately add the uploaded item to your list of foods.


davidhealey:
Nice one

PR #666, opened 2023-01-08
Stay in upload editor when upload to Open Food Facts failed

EmilJunker:
I was just testing the new image upload feature to OFF, and then I noticed that the OFF website is currently down, so of course the upload failed. But then I realized that it would be beneficial to stay in the upload editor when the upload failed, so I don't lose the pictures and don't need to enter all the information again.


davidhealey:
Very good idea, I'm glad OFF was down when you were testing otherwise this would have been missed :)

Issue #532, opened 2022-03-27
Spinner doesn't stop when OFF isn't reachable


h-h-h-h:
At time of this writing, openfoodfacts.org or subdomains of it aren't reachable (browser says [in German]: "network timeout"). This causes Waistline to not stop spinning its spinner after scanning a barcode. There should probably be a timeout. That timeout shouldn't be too long in my opinion. But if you think longer waiting than, say, 5 or 10 seconds could make a difference, maybe show a floating button at the bottom for a short amount of time that allows you to continue with a longer timeout (like the "Switch to" button in Firefox for Android after opening a new tab in the background). Or even a full dialog that informs the user and that probably fits better with uninterrupted continuation of trying to establish a connection.

Whether OS back button cancels the spinner or there's the same problem with a persisting spinner like it was the case with the backup feature was inconsistent for me; I experienced both behaviors. Maybe it had to do with switching the active app.

Proof from https://downforeveryoneorjustme.com/openfoodfacts.org:

![openfoodfacts org down](https://user-images.githubusercontent.com/13482553/160263441-f654bffb-6592-4755-a4e9-2d7a0ac9c27a.png)

EmilJunker:
@sethidden Thank you, this should be fixed now.

PR #667, opened 2023-01-08
Fix online search when result is undefined

EmilJunker:
This fixes another issue I just noticed because OFF is currently down.


davidhealey:
While you're at it, did you see the latest comment on this one - https://github.com/davidhealey/waistline/issues/532#issuecomment-1374825679

EmilJunker:
> While you're at it, did you see the latest comment on this one - [#532 (comment)](https://github.com/davidhealey/waistline/issues/532#issuecomment-1374825679)

This PR fixes precisely that issue

PR #668, opened 2023-01-09
Check response status when sending HTTP requests

EmilJunker:
OFF was down again for a while today, and I noticed one more potential problem with our error handling when making HTTP requests. It should be fixed now.


davidhealey:
They've been down a lot recently it seems, but good opportunity to find things like this.

PR #672, opened 2023-01-21
Default to system theme

scottmmjackson:
On first install the application should default to the user's preference, only overriding it on configuration.

### DIFF
--- a/www/activities/settings/js/settings.js
+++ b/www/activities/settings/js/settings.js
@@ -629,7 +629,7 @@ app.Settings = {
         usda: false
       },
       appearance: {
-        mode: "light",
+        mode: "system",
         theme: "color-theme-red",
         animations: false,
         locale: "auto",


EmilJunker:
When I implemented the setting to follow the system theme, I was reluctant to make it the default because it only works on Android 11 (or 10?) upward, and I don't know what happens if you enable in on older versions of Android.

davidhealey:
I'd like to see this tested on an older version of Android and if necessary a fallback implemented.

scottmmjackson:
I can build an APK and test on an old AVD

scottmmjackson:
About to start up for a test but my prediction is that it will present as light for as long as "system preferred" is selected.

https://github.com/framework7io/framework7/blob/master/src/core/components/app/app-class.js#L167 (when `window.matchMedia` not found, the function does nothing.

scottmmjackson:
Learned: API level 24 doesn't work because Chromium lacks the `async` keyword 😅

scottmmjackson:
Works fine on API 25.

scottmmjackson:
There is a potential follow up here though. On API 25, if you change to "use system theme" from another theme, the behavior is unpredictable.

I stand behind this as a solid improvement, but I think my ideal world looks like this:
- If `matchMedia` is unavailable, the default should be `light` explicitly and `use system theme` should be disallowed. If on a subsequent startup `matchMedia` becomes available, nothing should change (arguably a one-time info message should appear?)
- Ostensibly it's possible that `matchMedia` could disappear- it would be horrible but Google has done worse. In this case, the _best_ UX would be somehow detecting the previous theme and keeping it, but I'm pretty sure that's made intentionally impactical. So instead, default to light theme.

EmilJunker:
It should be very straightforward to set the default theme to `system` only if `matchMedia` is available and to `light` otherwise.

Preventing the user from selecting `use system theme` if `matchMedia` is unavailable would require more work. Not only would you need to disable the option in the UI, but also prevent the user from setting it via a backup file import. I'm not sure if this is worth the effort for such a small number of affected users.

Handling the case that `matchMedia` would disappear seems overly cautious to me. Let's not make this more complicated than it needs to be :)

scottmmjackson:
Great, sounds like this is ready to move forward unless there's any feedback?

davidhealey:
This one ready to be merged?

EmilJunker:
As I wrote above, I would change it to
```js
mode: (window.matchMedia) ? "system" : "light",
```
so it uses the system theme by default only if it's supported by the device.

davidhealey:
> As I wrote above, I would change it to
> 
> ```js
> mode: (window.matchMedia) ? "system" : "light",
> ```
> 
> so it uses the system theme by default only if it's supported by the device.

@scottmmjackson Could you do this? and I'll get this merged.

PR #673, opened 2023-01-21
Change 'nutriments' to 'nutrients' in strings file and UI defaults

scottmmjackson:
I changed only the default and not the symbol names or strings file. Looking for feedback on whether that ought to be done to avoid confusion.

"Nutrients" is a far more common term referring to the various nutritional components of food. Nutriments is most often used to refer to the food itself (as in [this company](https://www.nutriment.co.uk/)), or in particular to the "four nutriments" in Buddhism which refer to contexts in which food is eaten. [Reference](https://plumvillage.org/library/sutras/discourse-on-the-four-kinds-of-nutriments).

### DIFF
--- a/www/activities/meals/views/meal-editor.html
+++ b/www/activities/meals/views/meal-editor.html
@@ -104,7 +104,7 @@
             <ul id="nutrition">
             </ul>
             <ul>
-              <li><a class="list-button" id="nutrition-button" data-localize="foods-meals-recipes.show-more-nutriments">Show more nutriments</a></li>
+              <li><a class="list-button" id="nutrition-button" data-localize="foods-meals-recipes.show-more-nutriments">Show more nutrients</a></li>
             </ul>
           </div>
         </li>
--- a/www/activities/recipes/views/recipe-editor.html
+++ b/www/activities/recipes/views/recipe-editor.html
@@ -135,7 +135,7 @@
             <ul id="nutrition">
             </ul>
             <ul>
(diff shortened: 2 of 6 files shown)


davidhealey:
Hi,

I chose nutriment specifically because not all nutrients are nutriments. For example fiber. We may also want to track other nutriments in the future such as water, food additives, colourings, etc.

scottmmjackson:
@davidhealey I'm not sure I understand the rationale here.

First, I'm not sure "not all nutrients are nutriments" is either true nor relevant, as "dietary fiber" is listed under "nutriments" and would therefore be incorrect.

I can only speak for American English but "nutriments" is essentially never used in this context- we use "nutrients".

It may be that we want to clarify the default locale and provide either `en_US` or `en_GB` as the case may be. As it stands it seems we're using American spelling for at least some words:

https://github.com/davidhealey/waistline/blob/master/www/assets/locales/locale-en.json#L9

davidhealey:
> "dietary fiber" is listed under "nutriments" and would therefore be incorrect.

Fiber is a nutriment, it is not a nutrient. It isn't digested and doesn't provide nutrition.

davidhealey:
I'll leave this for now. If there is some demand from more users about this I'll revisit it.

EmilJunker:
If I may add my two cents here, as a non-native speaker I also feel like nutriments sounds a bit odd in this context. I would agree with the sentiment that the term nutrients is more common, but I don't have very strong feelings about this.

davidhealey:
Hmm, well I guess if the change is only to the front-end anyway then it doesn't really affect anything internally, so why not.

PR #699, opened 2023-02-23
Nutrition Name Corrections - English

daygle:
Ensure that nutrition labels are shown with a capital letter for the first character of each word. Also updated the word 'Proteins' to 'Protein' as shown on nutrition labels.


davidhealey:
I think this needs to be done through crowdin

daygle:
> I think this needs to be done through crowdin

Really? A assumed source code was maintained here?

davidhealey:
From the readme

image

davidhealey:
I think the en one might not be included though. I'll need to check.

daygle:
> I think the en one might not be included though. I'll need to check.

Thanks David

davidhealey:
Ok from what I can tell the en is like the template file that all the others read from. So hopefully if I merge your change it won't be overwritten by Crowdin.

daygle:
Thank you :)

PR #738, opened 2023-08-10
Fix camera and storage on Android 13+

### LINKED ISSUE #733: [BUG] Can't open camera on adding new element
Hello,

i have a Pixel 6a with latest Android 13 update.
I can't add a picture on adding new element/food, the UI say that "There was a problem using the camera"
If i use camera to scan a bar code it work, so It doesn't look like a permissions issue

EmilJunker:
This fixes the permission issues with Android 13/SDK 33 and closes #733.

I replaced the official `cordova-plugin-camera` with the fork from [felicienfrancois/cordova-plugin-camera](https://github.com/felicienfrancois/cordova-plugin-camera). This one is compatible with Android 13. Once the official plugin is fixed, we can switch back to that one.

I also updated the files plugin and changed the logic for the backup. If the user is on Android 13 or above, the backup files are now put in: `Android/data/com.waist.line/files`

On older versions of Android, it continues to use the backup location in the Documents folder: `Documents/Waistline/1667758046`

For users on Android 13 or above, this has the disadvantage that the backup files will be deleted when the app is uninstalled. But I don't see any way to prevent this. It looks like apps are no longer allowed to write to the Documents folder on Android 13 ☹️ 

Note that because of the changed plugins, you need to install the new versions before you build. The safest way to do that seems to be to delete the `platforms` and `plugins` folders, do `npm install`, and then re-add all the platforms with `cordova platform add <platform-name>`. I don't know why this stuff always has to be complicated 😅

### DIFF
--- a/metadata/en-US/changelogs/30601.txt
+++ b/metadata/en-US/changelogs/30601.txt
@@ -1,7 +1,0 @@
-- added language options for Czech and Romanian
-- added new splashscreen and app icon, including monochrome support
-- the statistics page now remembers both the selected stat and range
-- various bugfixes
-
-You can find the full changelog at 
-https://github.com/davidhealey/waistline/releases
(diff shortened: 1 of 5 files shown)


davidhealey:
Thanks Emil!

Issue #758, opened 2023-09-30
Feature Request: Option to show/hide Brands in the Diary screen


hellsend:
On small-ish screen size, with large-ish Android text size, the diary view gets a little bit clogged up for my key-user taste. 

I suggest we add the option to show or hide the food/meal/recipe item brand in the diary screen, so that users who so wish can do away with this information on the Diary screen. 
This is only a Diary rendering modification, no database or "hard-structure" modification should be needed. 

From what I've seen in the code, the implementation should be quite simple: 

- Define the option for the first-time setup, 
- Add the option in the settings/diary html page, 
- Add the parameter in the foods/meals/recipe renderItem function and add the parameter condition on the Brand subtitle section, 
- Initialize the variable & use it when calling the renderItem function in the diary/groups js page, 
- Allow for translation of the Option in the settings. 

I've tried to do that already in the code, and local (browser-based) & personal apk (app-debug.apk installed on my phone) testing has been successful. 

I would like to create the pull request to implement these modifications.
 
I'm not sure about all of the translations, but I've tried to make them in line with the other "Show xxx" options in each localization file. 

Feel free to give me some feedback on this, it's my first pull request on this codebase (maybe even my first ever github pull request?) and I'd like not to break everything :-) 

Cheers,
Hellsend-

davidhealey:
Sounds good, let's see the code

hellsend:
Just did that, and I'm impressed with your reaction time :-D (I'll try and not hold you up to this kind of standard in the future, I promise :-))

davidhealey:
Looks good. I haven't tested it yet though.

The translations are handled through Crowdin, you only need to add strings directly to the locale-en file. Can you remove the other files from your PR? They will be overwritten by changes from Crowdin.

hellsend:
Oh, well, yeah, let me just try and find out how to remove files from a commit and I'll push that right away :-)

hellsend:
@davidhealey : Okay, after much (uneeded) suffering, and possibly some erasing of some modifications (comments?) made by @EmilJunker , I think the request #760 is more in line with what you mentioned. 

Feel free to test around, I didn't have any problems (built locally with Gradle 7.6.3) on my end, to try and respect the Dockerfile instructions).

PR #759, opened 2023-09-30
Add 'Show/Hide Brand' option in Settings/Diary and make it work on the Diary screen

hellsend:
As stated in the issue #758 , here is the related modifications to being able to show/hide the items brands on the diary screen. 

Again, I hope the work is alright and the locales are acceptable (I'm really not sure what happened with the -si one...). 

Don't hesitate to get back to me on this one! 

Cheers-


www/activities/settings/views/diary.html
```diff
@@ -39,6 +39,20 @@
           </a>
         </li>
 
+        <li>
+          <div class="item-content">
+            <div class="item-inner">
+              <div class="item-title" data-localize="settings.diary.brands">Show Brands</div>
+              <div class="item-after">
+                <label class="toggle toggle-init">
+                  <input type="checkbox" field="diary" name="brands" />
```
EmilJunker, line 48: The input name should be "show-brands" instead of "brands" to better describe what this setting does. It also needs to match the property name in firstTimeSetup (which it currently doesn't).

www/activities/settings/js/settings.js
```diff
@@ -613,6 +613,7 @@ app.Settings = {
       diary: {
         "meal-names": ["Breakfast", "Lunch", "Dinner", "Snacks", "", "", ""],
         timestamps: false,
+        "show-brand": false,
```
EmilJunker, line 616: The property name here needs to match the one used elsewhere in the code. I suggest changing it to "show-brands".

Also, I think this setting should be true by default because most users will want to see the brand.

Finally, you also need to add this setting to the migrateSettings function (further below in this file) and set it to true there to make sure that users who already have the app installed and upgrade to the new version don't suddenly have their brands disappear from the diary.

www/activities/foods-meals-recipes/js/foods-meals-recipes.js
```diff
@@ -421,7 +421,7 @@ app.FoodsMealsRecipes = {
     return 0;
   },
 
-  renderItem: async function(data, el, checkbox, sortable, clickable, clickCallback, tapholdCallback, checkboxCallback, timestamp, thumbnailSetting) {
+  renderItem: async function(data, el, checkbox, sortable, clickable, clickCallback, tapholdCallback, checkboxCallback, timestamp, thumbnailSetting, showBrand) {
```
EmilJunker, line 424: The renderItem function is called in many places throughout the app, not just from the diary. To make sure the behavior doesn't change in these other places, you need to pass `true` for the showBrand parameter there.

Also I know this is nitpicking, but I think it would be better to put the new showBrand parameter next to the existing timestamp parameter instead of at the end.

EmilJunker:
Thank you for the pull request. This will be a useful feature to have.

I left a few comment in the code. Please don't hesitate to ask if something is unclear.

hellsend:
Wait, sorry, I kinda struggled with the whole branch (because I'd modified locales files when I shouldn't have) and I've deleted, then re-created, the branch, so any modifications you might have made as well, well... I'm afraid they're lost...

The "new" issue758 branch should be ready for review. 
Again, sorry about that, I should have deleted this "bad" branch as soon as I had the input from David.

www/activities/settings/views/diary.html
hellsend: Now I see it, okay, thanks! 

I'll adapt this right away and _try to_ merge this.

www/activities/foods-meals-recipes/js/foods-meals-recipes.js
davidhealey: I think the number of parameters in this function is getting out of hand :) It will need to be refactored at some point so I don't suppose it matters too much what position this new parameter has. But if you prefer it next to timestamp I won't complain :)

www/activities/settings/views/diary.html
hellsend: @EmilJunker : Should be better now, my local testing seems to agree too.

www/activities/settings/js/settings.js
hellsend: Oh, sorry again, missed this. 
Will adjust my branch for the default value, and the migrateSettings.

www/activities/foods-meals-recipes/js/foods-meals-recipes.js
hellsend: @EmilJunker : Will also take this into account. 
@davidhealey : I probably won't try to refactor anything at this point, sorry ^_^"

hellsend:
@EmilJunker , @davidhealey : I suggest we continue the discussion on the pull #760 page ; I'll try and list all the changes there.

PR #763, opened 2023-10-01
Switch back to official camera plugin

### LINKED ISSUE #733: [BUG] Can't open camera on adding new element

EmilJunker:
A while ago I switched to a different camera plugin to fix some problems on Android 13. Now the official camera plugin from the Cordova team has been updated, so we can switch back to that one (closes #733).

After I switched back the plugin, I couldn't build at first because of conflicting storage access permissions in the AndroidManifest. I applied some fixes for this, and now the build should be working. Once this PR is merged, I would recommend following these steps to make sure the fixes are also applied in your setup:

1. Delete the `platforms` and `plugins` folders
2. Run `npm install`
3. Run `cordova platform add browser@7.0.0 && cordova platform add electron@3.1.0 && cordova platform add android@12.0.1`


davidhealey:
Thanks. I'll merge this when ready for another release and apply your suggested steps

PR #785, opened 2024-01-02
Add ability to read nutrition values aloud with Text to Speech (TTS)

### LINKED ISSUE #457: Read nutrition values with text-to-speech
After scanning a barcode, you need to check whether the nutrition values are up-to-date. Having to switch back and forth with your eyes between phone and food packaging is rather annoying; it's easy to lose track. If the bare nutrition values as they're presented without additional text could just be read out loud with TTS one after the other, this would allow for quick, hassle-free verification of the numbers.

EDIT: In another use case, I manually create a food and want to type in the numbers as fast as possible without thinking and verifying. Normally, this would be too inaccurate. But this all changes with this TTS feature. However, for this use case, it is important that the numbers that are passed on to the TTS engine are formatted in advance (at least in the background just for TTS), because, otherwise, my German TTS voice, that expects `,` as decimal separator and leading `0` before it, would try to read text like `.58`.

For this, a loudspeaker button in the nutrition section header would be perfect (otherwise at the bottom of the section?). Once tapped, the reading of all values begins. Tapping it while reading is in progress would pause (continue before interrupted item). Long-pressing it would abort the reading. EDIT: If you don't want to bury related options in the settings, long-pressing could instead also reveal a context menu with items: Restart (has additional delay, if already actively reading), Options. Options are (possibly; maybe use defaults for the one or the other): Initial Delay, Rate (= speed), Silence Duration Between Items, (Voice?), (Pitch?). [SSML](https://en.wikipedia.org/wiki/Speech_Synthesis_Markup_Language) can be used for this (Web Speech API at least supports it). Rough settings like just an extra delay boost might be enough. If reading fails, show a toast.

EmilJunker:
This PR closes #457.

#### Added button in the Food Editor:

Screenshot_20240102-123315

#### Added settings page:

Screenshot_20240102-123226

The TTS functionality is based on the plugin [`cordova-plugin-tts-advanced`](https://github.com/spasma/cordova-plugin-tts-advanced). It's all pretty straightforward.
I'm not sure how useful this feature will really be, but I think it certainly has its use cases, and I had fun implementing it :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "cordova-plugin-inappbrowser": "^5.0.0",
     "cordova-plugin-ionic-keyboard": "2.2.0",
     "cordova-plugin-network-information": "^3.0.0",
+    "cordova-plugin-tts-advanced": "^0.5.3",
     "patch-package": "^7.0.0",
     "phonegap-plugin-barcodescanner": "^8.1.0"
   },
@@ -36,6 +37,7 @@
       "cordova-plugin-inappbrowser": {},
       "cordova-plugin-ionic-keyboard": {},
       "cordova-plugin-network-information": {},
+      "cordova-plugin-tts-advanced": {},
       "phonegap-plugin-barcodescanner": {
         "ANDROID_SUPPORT_V4_VERSION": "27.+"
       }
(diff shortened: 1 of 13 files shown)


davidhealey:
That looks like a great little addition, I'm sure lots of people will find it useful!

PR #788, opened 2024-01-12
Fix OFF search error handling

### LINKED ISSUE #787: Barcode scanner hangs when OpenFoodFacts is down for maintenance
I know this is quite a small issue, and wouldn't happen a lot. I just spent 10 minutes playing around with my phone's internet connection, clearing app cache etc., before I realised what was happening.

Might I suggest a timeout?

https://github.com/davidhealey/waistline/assets/143519476/a0517f01-80d4-4b1b-bebd-850eb423d699

EmilJunker:
This fixes #787.

@Dermanthi Thank you for the bug report. The app was actually already using a timeout of ten seconds when making requests. But there was a problem with the internal error handling.

### DIFF
--- a/www/activities/foodlist/js/foodlist.js
+++ b/www/activities/foodlist/js/foodlist.js
@@ -448,7 +448,7 @@ app.Foodlist = {
 
                 if (result !== undefined && result[0] !== undefined) {
                   item = result[0];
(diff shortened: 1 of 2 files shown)


Dermanthi:
My mistake with the assumption. Thank you.

PR #816, opened 2024-05-02
Add ability to take photos for meals and recipes

### LINKED ISSUE #811: Locally Changing OFF pics
Good evening, 

I was wondering if we can have the option to be able to change the pictures in which people upload to OFF. The idea here is that it would be local only as I know we can't make edits to existing items within the app. Sometimes people post pictures and it can be incorrect or has a hand in it, etc. I personally prefer stock photos of an item. Having this option and ability I believe will help us with cataloging our food items further if we so choose. Thank you for listening.

Regards

### LINKED ISSUE #812: Adding pictures to meals and recipes
To help with visual cataloging, I think it would be helpful to have the option to be able to add our own pictures to meals and recipes as well as having the option to automatically utilize the pictures of the various items that make up the meal or recipe. For example, if i have a recipe for rice balls, it could create something like the below. 
Screenshot_20240423-203835
These would just be additional options to add and not need to be default. Let me know your thoughts. Thanks!

EmilJunker:
This adds the ability to take photos of meals and recipes. Just like the existing custom food images, they are stored locally in the app as base64 strings. Closes #812.

In addition, it is now possible to remove pictures imported from OFF and replace them with local images. Closes #811.

**Note: This PR will likely have a small merge conflict with #817. Whichever PR gets merged first, I will fix the conflict in the other one, and then it can also be merged.**

Screenshot_20240502_093000_Waistline Screenshot_20240502_092933_Waistline

Screenshot_20240502_093149_Waistline Screenshot_20240502_093141_Waistline

### DIFF
--- a/www/activities/foods-meals-recipes/js/food-editor.js
+++ b/www/activities/foods-meals-recipes/js/food-editor.js
@@ -474,6 +474,7 @@ app.FoodEditor = {
     if (item.image_url !== undefined && item.image_url !== "" && item.image_url !== "undefined") {
 
       if (app.Settings.get("foodlist", "show-images")) {
+        let removable = (app.FoodEditor.origin == "foodlist");
 
         if (item.image_url.startsWith("http")) {
           let wifiOnly = app.Settings.get("foodlist", "wifi-images");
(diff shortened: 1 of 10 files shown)


davidhealey:
Thanks! I'll merge 816 and 817 first

Issue #799, opened 2024-03-14
A way to add generic foods (like a fruit or an egg) without personal API key for USDA database


prototyperspective:
I think currently the biggest shortcoming of the app is that is doesn't support generic foods. This probably puts most users off before they learned the basics of the app. Prior issues about this: #222 #86 #154

The current way this has been addressed requires the user to somehow learn that and how it's possible and to create and enter a personal API key for the USDA food database. This is not a solution for the following reasons:
* Privacy (due to the personal API key)
* Most people don't know this and aren't able to find out (also [not even the FAQ](https://github.com/davidhealey/waistline/blob/master/FAQ.md#how-can-i-search-for-foods-online) has the info on how)
* …instead they rather turn to another food tracking app which supports generic foods, notably MyFitnessPal, or stop using the app altogether ([example](https://www.reddit.com/r/fossdroid/comments/f1nj39/food_tracking_app_how_to_use_waistline/))
* The effort and time required to set this up
* When searching it's still hard to find the generic item like "carrot" since it's not shown at the top
* That this is not an accessible way, 95% of users won't do this even if they knew this is somehow possible and very many of those that would try don't have the skills to do this (many don't even have a computer); it's not user-friendly and greatly reduces the usefulness of the app

Please add a way to add generic foods without a personal API key for the [FoodData Central database](https://fdc.nal.usda.gov/) by the US Departments of Agriculture. These are the potential solutions I have thought of so far (maybe there are more):
* import their data into any a database of yours (the data is licensed CC0)
* import their data into Open Food Facts
* use another database that doesn't require a personal API key
* get in contact with them so that a personal API key is not required 
* import their data into Wikidata or otherwise enter the data into Wikidata and then get the data from there (see my proposal [here](https://www.wikidata.org/wiki/Wikidata_talk:WikiProject_Food#Large-scale_imports_of_food_items_data_from_FoodData_Central_&_OpenFoodFacts_&_studies))
* identify which of the foods in Open Food Facts could be good generic items for the particular food and then show that (using the unique ID) at the top of the search results
**usually there already is some generic food somewhere in the results, it's just very hard to find and sometimes what one thinks is the generic food item is actually some drink with that name or something like that
[...]

PR #841, opened 2024-07-11
Idea on how to improve raw ingridients search

voronaam:
This is an idea on how to improve raw ingredients search when user has both OFF and USDA integrations configured after the conversation in issue #799 .

The idea here is to use OFF only for the barcode scans and any search by product name will only return USDA results.

I do not know how to test it, but the change seems simple enough to me. Feel free to treat this PR as a simple idea that you may certainly improve upon.

### DIFF
--- a/www/activities/foodlist/js/open-food-facts.js
+++ b/www/activities/foodlist/js/open-food-facts.js
@@ -23,10 +23,15 @@ app.OpenFoodFacts = {
       //Build search string
       let url;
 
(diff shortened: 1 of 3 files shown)


davidhealey:
How does this handle the scenario where USDA is disabled? I think the OFF only option barcode scans should only be available is USDA is enabled.

voronaam:
Updated the PR to only respect this option when USDA is enabled.

Ideally the toggle in the settings should be disabled, but I do not know how to do that.

This PR is mostly to give you an idea.

PR #853, opened 2024-09-04
Support browser dev

aspensmonster:
Howdy folks,

I'm currently enrolled in a Diet and Nutrition class, and plan on using Waistline for tracking my diet. I have some prior development experience, and the code base looks clean enough to contribute to without hassle :) That being said, I find it easier to develop the app in the `browser` platform than deal with `adb` and `Chromium` `WebView` debugging. In the process of getting up to speed, I noticed that CORS blocks requests to USDA and OFF by default. This PR introduces a config option for the `browser` platform that sets a CORS proxy. I use this one: `docker.io/imjacobclark/cors-container:latest`. With that proxy running and this PR, I'm able to navigate through the app without issues.

Let me know what y'all think. Is there a better or different option for developing locally?

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -47,4 +47,5 @@
     <preference name="FadeSplashScreenDuration" value="500" />
     <preference name="SplashMaintainAspectRatio" value="true" />
     <hook src="hooks/read_app_version.js" type="after_prepare" />
+    <hook src="hooks/set_cors_proxy.js" type="after_prepare" />
 </widget>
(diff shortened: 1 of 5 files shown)


EmilJunker:
I think it would be much easier to just disable CORS in the browser you use to debug Waistline. For any Chromium browser, you can create a special shortcut to launch it with disabled security by passing these flags:

```
--disable-web-security --user-data-dir=<some-local-directory>
```

aspensmonster:
Thanks for the heads up, @EmilJunker! It doesn't look like there is an equivalent option to `--disable-web-security` for Firefox to launch without CORS, but there *is* an add-on that appears to do the trick:

https://addons.mozilla.org/en-US/firefox/addon/cors-unblock/

I'll use that add-on for getting around CORS while working in the browser :) Closing this PR now.

PR #868, opened 2024-11-08
Ignore leading zeroes when performing barcode search against local DB

### LINKED ISSUE #701: (bug) not matching Barcodes with leading zeros
If I scan products like https://openfoodfacts.org/product/0031146262571 (or other barcodes starting with (multible) 0) Waistline asks every time to import the food, but never adding it to the dairy or selecting the existing Entry.
May a string/int casting issue while scanning?

### LINKED ISSUE #867: [BUG] Barcode scanner does not find matches in local database when UPC begins in zero
Sometimes the values returned in a scan are inaccurate, and I have to correct them when saving to my local database. However, after saving locally, scanning still only searches the internet. It would be nice to be able to match the locally-saved values with matching UPC codes.

terminalmage:
This allows for barcode scans to match local DB entries with multiple leading zeroes.

Fixes #867.
Fixes #701.

NOTE: This might not perform as well on very large DBs, since it no longer passes a query to `getAll()`.

### DIFF
--- a/www/assets/js/db-handler.js
+++ b/www/assets/js/db-handler.js
@@ -254,7 +254,7 @@ var dbHandler = {
                 items: [],
               };
 
-              // Add weight 
+              // Add weight
               let weight = weights[index];
 
               if (weight !== undefined) {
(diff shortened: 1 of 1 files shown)


EmilJunker:
A more efficient solution would be to remove all leading zeros from the barcode before storing items in the local  database in the first place. Then we could pass the scanned barcode (also with stripped leading zeros) to the `getAll` method to efficiently find the item.

However, this approach would require us to remove the leading zeros for all existing items in the database when users upgrade to the new version (for backwards compatibility). There already exists an `upgradeDatabase` function in the codebase for purposes like this, so implementing this shouldn't be too difficult.

davidhealey:
> A more efficient solution would be to remove all leading zeros from the barcode

Yeah I think this is the way to go if we can't just get it to work with leading zeros.

PR #869, opened 2024-11-09
WIP: Health connect support

samtate:
Initial development of adding Health Connect Support to the app. Sharing this first stages of development to let maintainers review and make sure they're happy with it so far before I do a lot of work on the feature.

So far I have added the health connect plugin and updated the build environment to the versions required to support Health Connect. I have made a rudimentary settings page where it can be enabled or disabled. I have tested I can send data to Health Connect

Next will be to send your diary data to the API when you enable it, and re-send more when more food is added.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -15,20 +15,26 @@
     <allow-navigation href="*" />
     <allow-intent href="market:*" />
     <platform name="android">
-        <preference name="android-minSdkVersion" value="21" />
-        <preference name="android-compileSdkVersion" value="33" />
-        <preference name="android-targetSdkVersion" value="33" />
+        <preference name="android-minSdkVersion" value="26" />
+        <preference name="android-compileSdkVersion" value="34" />
+        <preference name="android-targetSdkVersion" value="34" />
         <preference name="AndroidPersistentFileLocation" value="Compatibility" />
         <preference name="AndroidExtraFilesystems" value="files,sdcard,cache" />
         <preference name="AndroidInsecureFileModeEnabled" value="true" />
         <preference name="AndroidXEnabled" value="true" />
         <preference name="GradlePluginKotlinEnabled" value="true" />
+        <preference name="GradleVersion" value="8.4" />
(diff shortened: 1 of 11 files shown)


davidhealey:
> updated the build environment to the versions required to support Health Connect

Does this break any of the existing features or deprecate some devices?

samtate:
> Does this break any of the existing features or deprecate some devices?

I wouldn't have thought it would break existing features and everything still seems to work fine.

I have had to update the minSdkVersion from 21 to 26: I understand this means that Android 5-7 are no longer supported? Android 8+ still consists of 96% of global Android devices and likely a higher percentage of users of this app.

samtate:
Hi @davidhealey @EmilJunker I'm keen to continue adding this but I am just wondering the best way to iterate through every food item stored in the database and get the dateTime, calories etc. Obviously everything only needs to be synced on the first enable of Health Connect: subsequent foods can be added to Health Connect when they are added to the diary.

Health Connect returns a unique ID when each food is added which can then be used to delete it later. Should this ID be stored alongside the other data for the food item? Then the delete from diary function can also call the delete from health connect function. 

An alternative is for Waistline to periodically delete all waistline health connect nutrition data and re-add from the current all-time diary. Not sure how this impacts performance on-device if it ran e.g. once a day.

davidhealey:
> Health Connect returns a unique ID when each food is added which can then be used to delete it later. Should this ID be stored alongside the other data for the food item? Then the delete from diary function can also call the delete from health connect function.
> 
> An alternative is for Waistline to periodically delete all waistline health connect nutrition data and re-add from the current all-time diary. Not sure how this impacts performance on-device if it ran e.g. once a day.

What about having a separate table for health connect data that contains both the health connect item id and the food id? It's been ages since I did any work myself on Waistline so I'm a bit out of the loop with the data storage. With SQL it's easy to setup an index with a cascade so that when an item is removed from one table it automatically removes it from the other, I don't know if a similar thing is possible with indexeddb...

samtate:
Thanks for the pointers. I will try and look into the best way to do it.

PR #875, opened 2024-11-24
Fix handling of barcodes with multiple leading zeros

### LINKED ISSUE #701: (bug) not matching Barcodes with leading zeros

### LINKED ISSUE #867: [BUG] Barcode scanner does not find matches in local database when UPC begins in zero

EmilJunker:
This implements the suggestion I made in https://github.com/davidhealey/waistline/pull/868#issuecomment-2465065746 and also addresses the concerns raised in https://github.com/davidhealey/waistline/pull/868#issuecomment-2465490929 by keeping the original barcode (with all leading zeros) for display purposes.

Fixes #701.
Fixes #867.
Closes #868.

### DIFF
--- a/www/assets/js/db-handler.js
+++ b/www/assets/js/db-handler.js
@@ -24,7 +24,7 @@ var dbHandler = {
     return new Promise(async function(resolve, reject) {
       //Open database
       var databaseName = 'waistlineDb';
-      var databaseVersion = 33;
+      var databaseVersion = 34;
       var openRequest = indexedDB.open(databaseName, databaseVersion);
 
       //Error handler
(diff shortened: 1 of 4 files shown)


www/assets/js/db-handler.js
```diff
@@ -174,6 +174,36 @@ var dbHandler = {
...
+
+            if (cursor) {
+              let value = cursor.value;
+
+              if (value.barcode !== undefined) {
+                if (value.barcode.startsWith("0")) {
+                  value.originalBarcode = value.barcode;
+                  value.barcode = value.barcode.substring(1);
+                }
+                if (value.barcode.startsWith("fdcId_")) {
```
davidhealey, line 193: I believe the "fdcId" prefix is to avoid clashes with barcodes from other systems such as open food facts - although it's been so long since it was written that I can't remember for certain. Do you think we'll run into any issues by removing the prefix?

EmilJunker: I only remove it for the `originalBarcode` property which is used for displaying and searching. The `barcode` property still has the prefix.

davidhealey: Ah ok, gotcha! Time for a new release, or do you have some more commits to make first?

EmilJunker: In short, items from OFF will be stored like this:

```
barcode: "<barcode-without-leading-zero>"
originalBarcode: "<barcode-including-leading-zero>"
```

And items from USDA will be stored like this:

```
barcode: "fdcId_<id>"
originalBarcode: "<id>"
```

EmilJunker: > Time for a new release, or do you have some more commits to make first?

I'm done for the time being, you can go ahead with the release 🚀

davidhealey: Thanks as always

davidhealey:
It's been ages since I compiled so I'm setting up a new VM for it, can you remember which version of Cordova Android I should be using?

EmilJunker:
@davidhealey It should be cordova 12.0.0 and cordova-android 12.0.1

PR #884, opened 2024-12-12
Add toggleable macro elements summary in category footer to see macro…

mgrzaslewicz:
… elements per meal easily

# Before
I found it hard to see how much protein and other macros I should add to particular meal (category). I could achieve that by clicking on calories button and see category nutriments. Then after closing the dialog I could already forget macros values. And it required way too many clicks.

# After
There is:
- new Diary setting called 'Show macros summary per category'. Turned off by default
- macro nutriments summary next to calories, e.g. 'P15 / F20/ C52 400 kcal'. Prefixes stand for protein, fat, carbohydrate.
- displaying macro nutriments summary is aligned with goal displaying in diary footer. e.g. if fat is not displayed in the diary footer, it won't be displayed in category summary

image

image

image

I know there are missing translations, but I don't want to invest too much upfront in this pull request without knowing the likelihood of merging it.

### DIFF
--- a/www/activities/diary/js/group.js
+++ b/www/activities/diary/js/group.js
@@ -151,7 +151,7 @@ app.Group = {
     if (grams === undefined || isNaN(grams)) {
       return null;
     } else {
-      const nutrimentPrefix = nutriment.charAt(0).toUpperCase();
+      const nutrimentPrefix = app.strings["nutriments-abbreviations"][nutriment] || nutriment.charAt(0).toUpperCase();
       return nutrimentPrefix + app.Utils.tidyNumber(Math.round(grams));
     }
   },
--- a/www/activities/diary/js/group.js
+++ b/www/activities/diary/js/group.js
@@ -130,10 +130,10 @@ app.Group = {
     });
 
     let categoryMacroNutriments = app.Settings.get("diary", "show-macro-nutriments-summary") ? [
-      this.getMacroNutrimentFooterText('F', nutrition.fat, categoryEnergyTotal),
-      this.getMacroNutrimentFooterText('C', nutrition.carbohydrates, categoryEnergyTotal),
-      this.getMacroNutrimentFooterText('P', nutrition.proteins, categoryEnergyTotal)
-    ].filter(text => text !== '') : [];
+      this.getMacroNutrimentFooterText('fat', nutrition.fat, categoryEnergyTotal),
+      this.getMacroNutrimentFooterText('carbohydrates', nutrition.carbohydrates, categoryEnergyTotal),
+      this.getMacroNutrimentFooterText('proteins', nutrition.proteins, categoryEnergyTotal)
+    ].filter(text => text !== null) : [];
 
     right.innerText = categoryMacroNutriments.join(' / ') +
         (categoryMacroNutriments.length > 0 ? ' / ' : '') +
@@ -141,14 +141,18 @@ app.Group = {
     row.appendChild(right);
   },
 
-  getMacroNutrimentFooterText: function(prefix/*: string*/, grams/*: number | undefined*/, categoryEnergyTotal/*: number*/) {
+  getMacroNutrimentFooterText: function (nutriment/*: string*/, grams/*: number | undefined*/, categoryEnergyTotal/*: number*/) {
     if (categoryEnergyTotal === 0) {
-      return '';
+      return null;
+    }
+    if (!app.Goals.showInDiary(nutriment)) {
+      return null;
     }
     if (grams === undefined || isNaN(grams)) {
-      return '';
-    }  else {
-      return prefix  + app.Utils.tidyNumber(Math.round(grams));
+      return null;
+    } else {
+      const nutrimentPrefix = nutriment.charAt(0).toUpperCase();
+      return nutrimentPrefix + app.Utils.tidyNumber(Math.round(grams));
     }
   },
 
--- a/www/activities/diary/js/group.js
+++ b/www/activities/diary/js/group.js
@@ -148,7 +148,7 @@ app.Group = {
     if (grams === undefined || isNaN(grams)) {
       return '';
     }  else {
-      return prefix  + Math.round(grams);
+      return prefix  + app.Utils.tidyNumber(Math.round(grams));
     }
   },
 
--- a/www/activities/diary/js/group.js
+++ b/www/activities/diary/js/group.js
@@ -151,8 +151,8 @@ app.Group = {
     if (grams === undefined || isNaN(grams)) {
       return null;
     } else {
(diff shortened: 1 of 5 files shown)


davidhealey:
Do you think there is enough room to put the full name, "Protein", "Carbs", "Fat" rather than individual letters?

mgrzaslewicz:
> Do you think there is enough room to put the full name, "Protein", "Carbs", "Fat" rather than individual letters?

I don't think so
image

davidhealey:
Yeah with Carbs it looks ok, could we put the unit after it too?

mgrzaslewicz:
> Yeah with Carbs it looks ok, could we put the unit after it too?

see my edited comment about smaller devices

davidhealey:
Yeah you're right, any idea if other apps have this feature and how they solve the layout?

mgrzaslewicz:
> Yeah you're right, any idea if other apps have this feature and how they solve the layout?

RP diet has a meal (category) focus and you can adjust everything easily. That's what I miss in this project
image

davidhealey:
Ok let's work with your original idea, just the letters and no units.

mgrzaslewicz:
> Ok let's work with your original idea, just the letters and no units.

Ok. Anything else needed apart from translations and package.json to be moved out of PR scope?

davidhealey:
At first glance it all looks good to me.

www/activities/diary/js/group.js
```diff
@@ -116,23 +116,42 @@ app.Group = {
...
     right.addEventListener("click", function(e) {
       app.Diary.showCategoryNutriments(id, nutrition);
     });
 
-    right.innerText = app.Utils.tidyNumber(Math.round(value), energyUnitSymbol);
+    let macroElementsTexts = app.Settings.get("diary", "show-macro-elements-summary") ? [
+      this.getMacroElementFooterText('P', nutrition.proteins, categoryEnergyTotal),
+      this.getMacroElementFooterText('F', nutrition.fat, categoryEnergyTotal),
+      this.getMacroElementFooterText('C', nutrition.carbohydrates, categoryEnergyTotal)
+    ].filter(text => text !== '') : [];
```
EmilJunker, line 136: Please change the order to [fat, carbohydrates, proteins] because this is the order we also use in other places. See our global `energyMacroNutriments` list:

https://github.com/davidhealey/waistline/blob/70d5821de6c3462af1aac75f0e0f4e04a18d2472/www/index.js#L34

www/assets/locales/locale-pl.json
```diff
@@ -262,7 +262,8 @@
       "wifi-thumbnails": "Pokazuj miniatury tylko przez Wi-Fi",
       "show-all-nutriments": "Pokaż wszystkie składniki odżywcze na stronie podglądu",
       "show-units": "Pokaż jednostki odżywcze",
-      "prompt-add-items": "Pytaj o ilość podczas dodawania produktów"
+      "prompt-add-items": "Pytaj o ilość podczas dodawania produktów",
+      "show-macro-elements-summary": "Pokazuj podsumowanie makroelementów per kategoria"
```
EmilJunker, line 266: Do not add translations as part of the pull request. Only add the new locale strings to the `locale-en.json` file. Other languages will be handeled later via Crowdin.

www/activities/diary/js/group.js
```diff
@@ -116,23 +116,42 @@ app.Group = {
...
   },
 
+  getMacroElementFooterText: function(prefix/*: string*/, grams/*: number | undefined*/, categoryEnergyTotal/*: number*/) {
+    if (categoryEnergyTotal === 0) {
+      return '';
+    }
+    if (grams === undefined || isNaN(grams)) {
+      return '';
+    }  else {
+      return prefix  + Math.round(grams);
```
EmilJunker, line 151: Please use `app.Utils.tidyNumber()` to format the gram values. That function does some important string magic to support LTR languages.

mgrzaslewicz:
I've improved the feature to align displaying category summary nutriments with goal settings as after using app for some time it just makes sense.

@EmilJunker @davidhealey Can I somehow help you to move forward with this pull request and release?

davidhealey:
Did you resolve all of @EmilJunker's comments? (looks like you did)

mgrzaslewicz:
> Did you resolve all of @EmilJunker's comments? (looks like you did)

Yes

www/activities/diary/js/group.js
```diff
@@ -130,25 +130,29 @@ app.Group = {
...
+    if (!app.Goals.showInDiary(nutriment)) {
+      return null;
     }
     if (grams === undefined || isNaN(grams)) {
-      return '';
-    }  else {
-      return prefix  + app.Utils.tidyNumber(Math.round(grams));
+      return null;
+    } else {
+      const nutrimentPrefix = nutriment.charAt(0).toUpperCase();
```
EmilJunker, line 154: Please make it possible to translate the prefixes into other languages. For example, carbohydrates are called "Kohlenhydrate" in German, so it should display the letter K in that case.

To do this, simply add the English abbreviations to the [`locale-en.json`](https://github.com/davidhealey/waistline/blob/master/www/assets/locales/locale-en.json) file, for example like this:

```diff
 {
   "nutriments": {
     ...
   },
+  "nutriments-abbreviations": {
+    "fat": "F",
+    "carbohydrates": "C",
+    "proteins": "P"
+  },
   "unit-symbols": {
     ...
   },
 }
```

Abbreviations in other languages can then be added later via Crowdin.

In the code, you can then get the localized prefix via the `app.strings` object:

```js
const nutrimentPrefix = app.strings["nutriments-abbreviations"][nutriment] || nutriment.charAt(0).toUpperCase();
```

Also, I think it might look better if instead of
`F9 / C13 / P6 / 150 kcal`
we display it as
`F 9 / C 13 / P 6 / 150 kcal`
or maybe even
`9 F / 13 C / 6 P / 150 kcal`

At least it would be worth exploring some of these alternatives...

EmilJunker: Also note that there could be languages where just using the first letter is not enough because the words for fat, carbs, and/or protein start with the same letter, so longer abbreviations are needed to distinguish them. Using the locale file and Crowdin makes that easy for us.

mgrzaslewicz: Done

davidhealey:
@EmilJunker @mgrzaslewicz Where are we at with this one? Ready to merge?

EmilJunker:
LGTM

mgrzaslewicz:
> @EmilJunker @mgrzaslewicz Where are we at with this one? Ready to merge?

I've been using apk with this change on my mobile for a few weeks already and works well

PR #906, opened 2025-04-06
Fix regression with floats

oxxur:
Entering floating point numbers now does not cause problems anymore.
This should fix the regression introduced in #905.

### DIFF
--- a/www/activities/foods-meals-recipes/views/food-editor.html
+++ b/www/activities/foods-meals-recipes/views/food-editor.html
@@ -133,7 +133,7 @@
                     <label for="portion" data-localize="food-editor.serving-size">Serving Size</label>
                   </div>
                   <div class="item-input-wrap">
-                    <input class="align-end auto-select" type="tel" pattern="[0-9]+[\+\-0-9]*" id="portion" name="portion" required validate>
+                    <input class="align-end auto-select" type="tel" pattern="[0-9]+[,\.\+\-0-9]*" id="portion" name="portion" required validate>
                   </div>
                 </div>
               </li>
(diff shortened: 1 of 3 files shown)


www/activities/foods-meals-recipes/views/food-editor.html
```diff
@@ -133,7 +133,7 @@
                     <label for="portion" data-localize="food-editor.serving-size">Serving Size</label>
                   </div>
                   <div class="item-input-wrap">
-                    <input class="align-end auto-select" type="tel" pattern="[0-9]+[\+\-0-9]*" id="portion" name="portion" required validate>
+                    <input class="align-end auto-select" type="tel" pattern="[0-9]+[,\.\+\-0-9]*" id="portion" name="portion" required validate>
```
davidhealey, line 136: Does `type=tel` work at forcing the correct keypad to display?

oxxur: That's the keypad I'm getting on my phone:

grafik
And with the symbols you get + and -:
grafik

davidhealey:
I think I prefer the type=number. I don't like having to go into a second menu to get the `.` I know you made it so it will convert `,` to `.` but that isn't obvious to the user. I don't see a downside to using number, at least on my phone it gives me `.`, `+`, and `-`

oxxur:
Interesting. I don't:
grafik


I get your point and I'd prefer having the cleaner layout as well, but that does not give me the + and -.

davidhealey:
Do none of the top bar buttons provide it?

oxxur:
No, none of the buttons provide anything other than they show.

davidhealey:
@EmilJunker What do you get on your phone when using the tel vs number input?

EmilJunker:
> @EmilJunker What do you get on your phone when using the tel vs number input?

Both tel and number keyboards look the same on my phone

Screenshot_20250407_161120_Edge

davidhealey:
I'm not 100% certain, but I guess we go with it then - until enough people complain, then we probably add a user preference :)

Issue #809, opened 2024-04-19
Automated math and calculations for meal planning


candroid-man:
There are spots in Waistline where with some simple automated math, could save users a ton of time. 

For example, I have a meal saved with chicken nuggets and Sriracha sauce. The serving size for the sauce is 36g, but what if I weighed out 42g this particular day? I wish I could just say how many grams I had and the app would adjust the nutrients according to the entered nutritional values for the product.

Also in a meal, if you use multiple servings of something it says 36g x 4 for example, so it would look like "36g x 4 (144g)". I would really like it if the app automatically did that math for you so when I'm measuring my food I can aim for the number in Waistline.

davidhealey:
> The serving size for the sauce is 36g, but what if I weighed out 42g this particular day? I wish I could just say how many grams I had and the app would adjust the nutrients according to the entered nutritional values for the product.

You can already do this. Either change the serving size once it's added to the diary or enable the "Prompt for quantity when adding items" setting in Settings >> Diary.

> I would really like it if the app automatically did that math for you so when I'm measuring my food I can aim for the number in Waistline.

I don't understand what you're asking for here.

candroid-man:
![ResizedImage_2024-04-19_18-45-25_8570.png](https://github.com/davidhealey/waistline/assets/51019003/eb7303c2-5d47-4c37-8743-ce566d9fc93b)

Here is a visual of what I was talking about. In parentheses it would be helpful for the total amount of grams instead of just 76g x 5

PR #916, opened 2025-07-16
Add total portion size of an item to the diary view

Frylen:
Display the total portion size (item.portion * item.quantity) of items in the diary view. See the following image for a few examples:

Implements feature request in issue #809.


davidhealey:
Looks good, you only need to add strings to the en locale, the rest are handled by Crowdin

PR #919, opened 2025-07-18
Improvement to linear regression trend

Frylen:
This is a small improvement on #918. 
#### Changes: 
- Instead of only showing a sign in the trend value when the value is negative, it will now display a + sign on a rising trend and a ± sign on a neutral trend. 
- The title of the trend field got updated from "Trend" to "Daily Trend" to show that the trend value is "per day" for the selected timeframe.
- Fixes the calculation, going from 0 to n-1 instead of 0 to n.


EmilJunker:
You are right, my mistake. Somehow my eyes deceived me. #918 indeed didn't contain any changes to the German translations, so this PR can be cleanly merged as it is.

PR #922, opened 2025-07-28
Allow takeover from all categories of the past day

oxxur:
This is to allow users to take over not only from the same category from the last day, but also from others.
The goal is to allow an easy takeover of leftovers of for example yesterday's dinner as today's lunch.

Since not everybody might want that I've made it optional and added a settings entry to control this feature. It is disabled by default.

### DIFF
--- a/www/activities/meals/js/meals.js
+++ b/www/activities/meals/js/meals.js
@@ -123,10 +123,26 @@ app.Meals = {
 
       // Get yesterday's meal if there is a diary category
       let category = app.FoodsMealsRecipes.getCategory();
-      let yesterdaysMeal = false;
 
-      if (category !== false) {
-        yesterdaysMeal = await app.Meals.getYesterdaysMeal(category);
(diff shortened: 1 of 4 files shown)


davidhealey:
This looks ok to me (I haven't tested it), can you remove the -de locale though, that will be handled by crowdin, you only need to provide the en one.

oxxur:
I have removed the -de location.

Here are some screenshots from the browser:
Settings:
grafik

When yesterday looked like this:
grafik

Takeover from yesterday with enabled option:
grafik


With Option disabled (and dinner selected):
grafik

oxxur:
Hi
I noticed there is a new release. Thanks for that.
May I ask why my PR did not make the cut?

davidhealey:
I didn't review any PRs when I made the release. It was made to get a [bug fix](https://github.com/davidhealey/waistline/pull/940) out. 

Looks fine to me though so I'll merge it now.

Issue #551, opened 2022-04-16
[Feature request] longer and/or arbitrary date range on statistics page


gottalovebrando:
Thank you for all the work you have done on this app! This is the best open source nutrient tracker I have seen!

I looked through the feature requests and didn't see this. It seems easy to implement (but I'm not an android programmer). I wanted to suggest the ability to select an arbitrary date range using a pop up calendar (already seems to be implemented on diary page). Or perhaps a little easier to code but not as useful to user -- ability to select longer time frames (2 year, 3 year, etc). The graph could also be made scrollable.

This would allow long term users of the app to see trends throughout time (or people who import data). Otherwise, it is essentially "lost" if no longer viewable.

I am using app version 3.4.0 on android 12.

davidhealey:
Good idea.

I'm not an Android programmer either ;)

Noxis73:
any updates on this? just came here to look for a feature request like that 😃

great app, btw! thank you!

davidhealey:
> any updates on this?

Not as far as I know

PR #953, opened 2026-04-25
Permit arbitrary date ranges in statistics view

brisk0:
This PR addresses Issue #551

Currently there are only fixed ranges for the statistics view, going up to a single year, and always ending at today.

This PR adds date selectors to the statistics screen to permit a user to enter a custom range, with any start and end dates.

The dropdown gets a new option, "Custom", and two date fields which use the native date picker and are constrained by one another to prevent "inverse" date ranges.

The dates are updated with any selection of the ranges in the dropdown, and are treated as the single source of truth for establishing the range in the statistics graph.

When the start and end date are changed, they set the dropdown field to "Custom".

Note that I haven't been able to get the android build to work, so I've only been able to test this using cordova web. Someone should definitely test this on Android first and confirm the date picker is acceptable.

20260425_192024_spectacle
20260425_191933_spectacle
20260425_192000_spectacle


davidhealey:
Code looks fine

EmilJunker:
Please add the English texts to the file under `www/assets/locales/locale-en.json`

brisk0:
Got an Android build, date picker looks pretty good to me
image

English texts have been added too.

PR #958, opened 2026-06-12
Add support to synchronize with intervals.icu

pierre-ernst:
Intervals.icu is an open platform to create workouts and track physical activities.
This PR adds an integration with Intervals.icu ("ICU") API.
Screenshot 2026-05-28 222637
Once enabled, users can add their API key and their own athlete ID
Screenshot 2026-05-28 222748
When the integration is enabled, a new button appears on the upper right corner, and allows 2 ways sync between Waistline and Intervals.icu
Screenshot 2026-06-09 225402

1. Calories spent during physical activities are pulled from ICU and displayed in the diary
2. Calories consumed are pushed to ICU and will show in the wellness data section of the corresponding day


davidhealey:
Is the workflow that the user has to manually sync each diary day one at a time? If so, is that what users will actually want?

In the commit I'm seeing a bunch of stuff that looks like nothing changed.

For the Icon use one included with FontAwesome rather than an external svg. See here for an example: https://github.com/davidhealey/waistline/blob/master/www/activities/diary/views/diary.html#L34

Probably also better to put the icon on the left side of the pie

pierre-ernst:
Thank you for your feedback @davidhealey 

1. I've updated the icon and moved it to the left
2. Sorry for the indentation changes, I did my best, but looks like previous commit were not consistent in terms of linting
3. I agree that a manual sync via the UI is not ideal, but given that Waistline relies on a local database, I think that's the best we can do
4. It's hard to answer 'is that what users will actually want?', but I think both Waistline and Intervals.icu cater to users who are interested in open data and tracking detailed metrics about their wellness/fitness. So Thanks for considering the suggestion.

davidhealey:
Thanks for making those changes. The noise on the indents is a bit horrible though. This might be something Claude could clean up if you pass it your repo.

pierre-ernst:
Done @davidhealey, the changes look better now

davidhealey:
Yeah that's good, thanks

PR #959, opened 2026-06-21
Add moving average option to statistics chart

### LINKED ISSUE #889: Add moving average for weight in Statistics page
I have checked for duplicate issues and I do not see any that match.

Due to day-to-day weight fluctuations, the graph of weight over time can be quite spiky. This makes it difficult to tell whether your weight is trending up or down at the planned rate for your bulk or cut. For example, a clean bulk might often be on the order of 0.25 lbs / week-- this is under typical amplitude of daily weight fluctuations so it is practically impossible to see whether you're on track with the current graph.

I think a smoothed out, less noisy moving average of weight would make the statistics page much more useful in this respect. I could see this being a trendline drawn on the same graph as the weight stat or a separate, calculated statistic.

I realize that the red average line exists, but I am requesting something different. The average across the last 3 or so months is not that helpful unless you are attempting to maintain your weight.

LuisMars:
Closes #889

Adds an optional moving average line to the statistics chart, smoothing the selected stat so it is easier to see whether it is trending up or down despite day-to-day noise. As raised in the issue thread, it applies to any stat, not just weight.

- "Show Moving Average" toggle and a configurable "Moving Average Period" (default 7) in Statistics settings
- Trailing simple moving average drawn as a second line over the chart, separate from the existing average annotation; ignores data gaps and works on both bar and line charts
- English strings added; other locales translated where possible, with English fallback

Tested in the browser build: Statistics settings -> enable Show Moving Average -> open Statistics -> pick a stat with history.


davidhealey:
This looks ok. The only thing I can see is the language files. You should only add your strings to the en file. All the others are handled through Crowdin.

LuisMars:
Fixed

PR #961, opened 2026-07-05
Guard Meals.init against overlapping calls

### LINKED ISSUE #942: Duplicate meals shown when adding to today’s dinner and wrong meal added
When adding a meal to today’s dinner, the picker sometimes shows duplicate entries for yesterday’s dinner with different calorie values. In other cases, the list shows the correct meals (e.g., “yesterday’s lunch” and “yesterday’s dinner”), but selecting one adds a different, seemingly random meal.

### Steps to reproduce

- Go to Diary → Dinner → Add meal.
- Observe the meal picker list.
- Select a meal from yesterday.

### Expected behavior

- Each previous meal appears only once.
- The selected meal is added to today’s dinner.

### Actual behavior

- Yesterday’s dinner appears twice with different calories.
- Sometimes the app adds a different meal than the one selected.

### Frequency
Intermittent. Happens often enough to be noticeable.

### Environment
Platform: Android 16
App version: 3.10.0
Device: Pixel 8

### Additional context
Screenshot attached showing duplicate “Yesterday’s Dinner” entries with different kcal values.

Image

munzzyy:
Fixes #942.

The Meals tab in the meal picker (Diary -> Dinner -> Add meal -> Meals) had no guard against `app.Meals.init()` running twice concurrently. If it gets triggered again before a previous call has finished - re-entering the Diary "Add meal" flow, switching tabs quickly, anything that fires `tab:init` again while the first `getListFromDB()`/`renderList()` pass is still awaiting IndexedDB - the two calls race each other. Each one clears and rebuilds the list independently, and depending on how the clears and appends interleave you end up with duplicate `<li>` rows in the DOM, including a duplicated "Yesterday's Dinner" row pulled from diary history. Since the checkbox selection is keyed off whatever ended up rendered in that slot, tapping one of the duplicates can add a different item than the one you actually picked.

I reproduced this locally in browser mode by seeding a diary entry for "yesterday" and calling `app.Meals.init()` twice without awaiting the first call - reliably got 2-3 duplicate rows in the DOM for a single logical item, matching the duplicate-entry behavior described in the issue. I didn't try to reproduce the differing-kcal detail from the screenshot specifically - that'd come from two independent `getTotalNutrition()` calls landing in the two interleaved render passes, which is consistent with a race but I didn't verify that exact path. A normal single visit to the tab, tab bouncing, and the search bar all still rendered a single copy of each item.

The fix: `init()` now stores its own in-flight promise. If it's called again while that promise is still pending, the second call just returns the same promise instead of kicking off a second fetch-and-render cycle. Once it settles the guard clears, so the next visit re-fetches normally - sequential re-inits were never the problem here, since `renderList(true)` already clears the list before repopulating it. It's only overlapping calls that race and leave duplicates behind.

I ran this against Cordova's browser platform (`cordova platform add browser`, `cordova run browser`) since I don't have an Android device set up - couldn't run it through `cordova build android` on my end. Didn't touch `Foodlist.js`, which has the same unguarded pattern, since it's outside what's reported here - happy to follow up separately if that's wanted.

### DIFF
--- a/www/activities/foodlist/js/foodlist.js
+++ b/www/activities/foodlist/js/foodlist.js
@@ -22,9 +22,24 @@ app.Foodlist = {
   list: [], //Main list of foods
   filterList: [], //Copy of the list for filtering
   isRendering: false,
+  initPromise: undefined, //Guards against overlapping init() calls duplicating list entries
   el: {}, //UI elements
 
-  init: async function(context) {
+  init: function(context) {
+    // If a previous init() is still in flight (e.g. the tab was re-entered before the
+    // last render finished), wait for it instead of starting a second, overlapping
+    // fetch/render cycle that would leave duplicate entries in the list.
+    if (app.Foodlist.initPromise !== undefined)
+      return app.Foodlist.initPromise;
+
+    app.Foodlist.initPromise = app.Foodlist.runInit(context).finally(() => {
+      app.Foodlist.initPromise = undefined;
+    });
+
+    return app.Foodlist.initPromise;
+  },
+
+  runInit: async function(context) {
 
     if (context && context.item) {
       await this.putItem(context.item);
--- a/www/activities/meals/js/meals.js
+++ b/www/activities/meals/js/meals.js
@@ -22,9 +22,24 @@ app.Meals = {
   list: [], //Main list of meals
   filterList: [], //Copy of the list for filtering
   isRendering: false,
+  initPromise: undefined, //Guards against overlapping init() calls duplicating list entries
   el: {}, //UI elements
 
-  init: async function(context) {
+  init: function(context) {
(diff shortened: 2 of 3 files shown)


EmilJunker:
Thanks, this looks good. But the init functions for the foodlist and for recipes work exactly the same as for meals, so they will have the same problem. It makes no sense to only fix this problem for the meals tab and not for the others. Please apply the fix to the recipes and foodlist as well.

munzzyy:
Fair point, done. Applied the same initPromise guard to Foodlist.init and Recipes.init - same shape as the meals fix, init() now just delegates to a runInit() and returns the in-flight promise if one's already running. Checked the syntax on all three files and wrote a small standalone script mimicking the guard to confirm overlapping calls collapse into one runInit while a later, non-overlapping call still re-fetches normally.

PR #964, opened 2026-07-05
Translate remaining untranslated strings in uk locale

munzzyy:
Fixes 5 strings in `locale-uk.json` that were still showing raw English: the USDA title, and four language names in the picker (czech, romanian, serbian-cyrillic, serbian-latin).

- settings.integration.usda-title -> Міністерство сільського господарства США
- locale.czech -> Чеська
- locale.romanian -> Румунська
- locale.serbian-cyrillic -> Сербська (Кирилиця)
- locale.serbian-latin -> Сербська (Латиниця)

The language names use the same adjective form as every other language already in that list (Болгарська, Голландська, Англійська), confirmed against the Ukrainian Wikipedia articles for each language (uk.wikipedia.org/wiki/Чеська_мова, /wiki/Румунська_мова, /wiki/Сербська_мова). The Serbian script variants match the pattern locale-ru.json already ships for the same keys (Сербский (Кириллица) / Сербский (Латиница)).

For the USDA title, used the name from Ukrainian Wikipedia's own article on the department, which is also what the U.S. Embassy in Ukraine uses on its own Ukrainian-language pages (ua.usembassy.gov).

Only touched these 5 strings, no other changes.


davidhealey:
Translations need to be made through [Crowdin](https://crowdin.com/project/waistline), any changes made in this repo will be overwritten.

PR #981, opened 2026-08-18
Fix Edge-To-Edge on Android 16

### LINKED ISSUE #977: Updating to 3.11.1 breaks fullscreen
Hi there, 

After installing the recent update from Google Play (3.11.1), all screens shows with less height than before. It is center on the screen, but there are empty spaces on the top and the bottom that weren't before (see green rectangles).

Image

Any ideas? I'm using a Samsung S21FE, running Android 16. Everything is stock.

Thanks!

EmilJunker:
I just updated the Dockerfile for Android 16 (SDK 36) and built the app. I can reproduce the layout problem described in #977 on my Samsung phone running Android 16.

But the change proposed in #978 doesn't fix the problem for me. I was able to fix it by adding the `AndroidEdgeToEdge` preference in config.xml. However, I don't know if this has any negative implications for devices still running older versions of Android. It would be good to test this before making a new release.

Here is the debug APK I built: [app-debug.zip](https://github.com/user-attachments/files/31174771/app-debug.zip)

Closes #977
Closes #982

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -18,6 +18,7 @@
         <preference name="android-minSdkVersion" value="21" />
         <preference name="android-compileSdkVersion" value="36" />
         <preference name="android-targetSdkVersion" value="36" />
+        <preference name="AndroidEdgeToEdge" value="true" />
         <preference name="AndroidPersistentFileLocation" value="Compatibility" />
         <preference name="AndroidExtraFilesystems" value="files,sdcard,cache" />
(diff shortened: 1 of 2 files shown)


EmilJunker:
@davidhealey You can merge this fix if you want. I have since confirmed that this change doesn't affect older versions of Android. Framework7 handles it all automatically.

davidhealey:
@EmilJunker Thanks for testing

Issue #902, opened 2025-03-03
[enhancement] diary collapsed by default


victorjagu:
Hello sirs,

Just wondering if it is possible to add an option to set the diary view as collapsed by default, or at least that remember the last setting.

When I ate a lot of different stuff for a single category (let say lunch), once I have finished adding meals and food I find the breakdown too overwhelming, and if it's collapsed I can see the total calories eaten, so that's perfect.

As a workaround, I'm currently using the "not show thumbnails" option, which helps somehow, but having it collapsed would be better to my needs.

Anyway, this is very personal, so just a kind request in case there is a little time to add this in the future.

Thanks!!

Issue #901, opened 2025-03-03
[enhancement] Mini Chart in Diary by Default


victorjagu:
Hello,

Just asking, if possible, for adding an option to set the mini chart visible in the diary as the default view.
I personally prefer that view because I identify better how much calories and macros I still have to eat.

Thanks!

PR #990, opened 2026-09-03
Add new default options for diary display

andyfranqueira:
Addresses #901, #902

**Issue**
User requested the ability to choose settings for the diary page to collapse groups and display mini charts by default.

**Solution**
Added new diary settings to support these options and updated the diary display to set the view based on these values.

**Testing**
- Ran the application using adb and a Pixel 8.
- Tested the new setting options and switched between options for a couple of days.
- The UI rendered as intended, was practical and readable, all other functionality unchanged.

**Settings**
901-settings

**Example**
901-example

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -98,6 +98,11 @@ app.Diary = {
 
     // Toggle nutrition swiper card
     if (!app.Diary.el.diaryNutrition.hasClickEvent) {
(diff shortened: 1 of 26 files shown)


davidhealey:
You only need to add strings in the en locale, the other languages are handled by crowdin

andyfranqueira:
Updated, I'm now off to read up on crowdin :)

Issue #891, opened 2025-01-16
Enhancement - Improve UX: Change "+" button to checkmark when items are selected


longestmt:
When making food selections I frequently tap the "+" button in the bottom right corner after making a selection. I think this is a learned behavior because it is intended to be the primary action for the user. Instead I have to select the checkmark in the top-right which is not intuitive.

My suggestion is this...

> 
> WHEN no food is selected THEN primary bottom-right button is "+"
>
> WHEN one or more food items are selected THEN primary bottom-right button is a check mark.
>

I think this user experience is more natural. The user can either add a new food when they can't find it listed or simply confirm their selections.

I'll provide a screenshot!

davidhealey:
Sounds good, I'm not sure how easily that is achieved though.

longestmt:
> Sounds good, I'm not sure how easily that is achieved though.

You would know better than me for sure 😂 That being said this would be a paradigm level change for the interaction so it improve the selection UX for meals etc as well. Hopefully it would make it simpler to replace the existing experience across the board instead of for these discrete scenarios.

PR #991, opened 2026-09-03
Add capability to switch the food selection button from add to select

andyfranqueira:
Addresses #891

**Issue**
User requested on the Food-Meals-Recipe page: when one or more food items are selected the primary bottom-right button should become a check mark and pivot to selecting rather than adding function.

**Solution**
Updated the action button to switch icons based on element selection and to execute the indicated functionality.

**Testing**
- Ran the application using adb and a Pixel 8.
- Tested the updated food selection user interface for a couple of days.
- The UI rendered as intended, the button was functional in both add and select modes, all other functionality unchanged.

**Before**
891-before

**After**
891-after-1
<br/>
891-after-2

### DIFF
--- a/www/activities/diary/js/diary.js
+++ b/www/activities/diary/js/diary.js
@@ -282,6 +282,10 @@ app.Diary = {
     let container = document.querySelector("#diary-day");
     container.innerHTML = "";
 
+    // Expand the target group's card when returning here after adding items to it
+    if (scrollPosition !== undefined && scrollPosition.category !== undefined && app.Diary.groups[scrollPosition.category] !== undefined)
+      app.Diary.groups[scrollPosition.category].collapsed = false;
+
     for (group in app.Diary.groups)
       await app.Diary.groups[group].render(container);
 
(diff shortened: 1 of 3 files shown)


davidhealey:
Does adding a new food when some are already selected clear the selection?

andyfranqueira:
> Does adding a new food when some are already selected clear the selection?

- When a food item has been selected the floating action button becomes a "use for diary" action. 
- So if a food item has been selected the user cannot add a new food.
- If the user deselects the food items, the floating action button reverts to an "add new item" action.

davidhealey:
I meant prior to your change

andyfranqueira:
> I meant prior to your change

Ah, I see.

Current flow:
- Select item 1
- Click add item
- Enter item 2
- Return to Foodlist
- Item 1 still selected
- Item 2 added but not selected

davidhealey:
I'm not sure which workflow would be better, the ability to add additional food while other are selected or having a more prominent check mark button. 

I'll leave this one for now and see if any other opinions are added.

andyfranqueira:
After trying this out for a few days I've noticed it creates inconsistency in the UI between accessing from the Diary and directly. If there is still desire to modify the select button it will need some deeper analysis and design.

